PR #194, opened 2026-04-16
Potential fix for 1 code quality finding

laboef1900:
_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/Compendiq/compendiq-ce/security/quality/ai-findings)._

PR #195, opened 2026-04-16
Potential fix for 1 code quality finding

laboef1900:
_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/Compendiq/compendiq-ce/security/quality/ai-findings)._

### DIFF
--- a/backend/src/domains/confluence/services/confluence-rate-limiter.ts
+++ b/backend/src/domains/confluence/services/confluence-rate-limiter.ts
@@ -106,9 +106,12 @@ export async function initRateLimiter(): Promise<void> {
       [SETTING_KEY],
     );
     if (result.rows.length > 0) {
-      const rpm = parseInt(result.rows[0]!.setting_value, 10);
-      if (!isNaN(rpm) && rpm > 0) {
-        setRateLimit(rpm);
+      const settingValue = result.rows[0]?.setting_value;
+      if (settingValue !== undefined) {
+        const rpm = parseInt(settingValue, 10);
+        if (!isNaN(rpm) && rpm > 0) {
+          setRateLimit(rpm);
+        }
       }
     }
   } catch (err) {


laboef1900 (commented):
## Automated PR Review

**PR**: #195 — Potential fix for 1 code quality finding
**Branch**: `ai-findings-autofix/...-confluence-rate-limiter.ts` → `dev`
**Scope**: `initRateLimiter()` in `backend/src/domains/confluence/services/confluence-rate-limiter.ts`

### Static Checks
| Check | Status |
|-------|--------|
| CodeQL | ✅ SUCCESS |
| Typecheck & Lint | ✅ SUCCESS |
| Tests | ✅ SUCCESS |
| Branch target (`dev`) | ✅ PASS |

### Code Review Findings

Diff replaces `result.rows[0]!.setting_value` (non-null assertion) with `result.rows[0]?.setting_value` + explicit `undefined` guard before `parseInt`.

- **Safety**: The outer `if (result.rows.length > 0)` already guarantees `rows[0]` exists, so the non-null assertion was technically sound. However, `setting_value` is a TEXT column that could theoretically be NULL — the previous code would have fed `null` into `parseInt`, yielding `NaN`, which the subsequent guard (`!isNaN(rpm) && rpm > 0`) would correctly reject. So the old code was safe-by-accident; the new code is safe-by-design.
- **Token-bucket semantics** unchanged — the diff touches only the startup-path DB read, not `acquireToken` / `refillTokens` / `resetRateLimiter`.
- **`CONFLUENCE_RATE_LIMIT_RPM` env var / admin UI override path**: unaffected; both still funnel through `setRateLimit(rpm)`.

### Coordination with PR #194

PR #194 modifies the test file and adds `await` to three `acquireToken()` calls in one test. That test mocks `query` to return `{ rows: [] }`, so it does **not** exercise the modified `initRateLimiter` path here. No overlap; merge order is irrelevant. Consider merging them together for cohesive history.

### Verdict: COMMENT

Fix is correct and strictly defensive. CI is green. Verdict capped at COMMENT (no local Docker smoke run in this review); recommend merging.

---
Reviewed by Claude | 2026-04-16

PR #202, opened 2026-04-16

laboef1900:
To fix this without changing functionality, replace each `new StorageEvent('storage', { key, newValue })` usage with a helper that creates a plain `Event('storage')` and defines `key` and `newValue` on it via `Object.defineProperty`. Then dispatch that event.

In `frontend/src/stores/auth-store.test.ts`:
- Add a small local helper inside the `describe` block (near setup hooks is best), e.g. `dispatchStorageEvent(key, newValue)`.
- Update the four affected test sections (lines around 82, 109, 128, 148) to call the helper instead of constructing `StorageEvent` with two args.
- No new imports or dependencies are required.

This removes constructor-signature ambiguity and keeps the tests’ intent unchanged.

### DIFF
--- a/frontend/src/stores/auth-store.test.ts
+++ b/frontend/src/stores/auth-store.test.ts
@@ -2,6 +2,13 @@ import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { useAuthStore } from './auth-store';
 
 describe('auth-store', () => {
+  const dispatchStorageEvent = (key: string, newValue: string | null) => {
+    const event = new Event('storage');
+    Object.defineProperty(event, 'key', { value: key });
+    Object.defineProperty(event, 'newValue', { value: newValue });
+    window.dispatchEvent(event);
+  };
+
   beforeEach(() => {
     localStorage.clear();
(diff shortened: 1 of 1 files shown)


laboef1900 (commented):
## PR Review: REQUEST_CHANGES

### Summary
Extracts a `dispatchStorageEvent()` helper that uses `new Event('storage')` + `Object.defineProperty` instead of the standard `StorageEvent` constructor.

### Issues

While the helper extraction reduces repetition (a minor positive), it's built on the same **false premise** as the other PRs in this batch: that `StorageEvent`'s second argument is superfluous. It is not — `StorageEventInit` is the standard way to set `key` and `newValue`.

**Same problems as sister PRs:**
- Produces a generic `Event`, not a `StorageEvent`
- Breaks `instanceof StorageEvent` checks
- Unnecessary complexity to work around a non-existent problem

### Verdict
**REQUEST_CHANGES** — Do not merge. If test DRY-ness is desired, a helper using the correct `new StorageEvent('storage', { key, newValue })` constructor would be welcome as a separate PR.

laboef1900:

Issue #258, opened 2026-04-20
follow-up(#256): preserve LLM queue + circuit breakers on call-site migration


laboef1900:
## Context

During execution of the Phase 5 call-site migration for issue #256 (multi-LLM-provider), a subagent surfaced two feature-parity gaps between the old \`providerStreamChat(userId, …)\` call path and the new \`resolveUsecase('chat') + streamChat(config, model, …)\` path:

### 1. LLM queue / concurrency cap (\`LLM_CONCURRENCY\` / \`LLM_MAX_QUEUE_DEPTH\`)

The old \`providerStreamChat\` in \`backend/src/domains/llm/services/llm-provider.ts\` wraps each call with \`enqueue(...)\` from \`./llm-queue.js\`, enforcing:
- \`LLM_CONCURRENCY\` (default 4 concurrent LLM requests)
- \`LLM_MAX_QUEUE_DEPTH\` (default 50, rejects when exceeded)

The new \`streamChat\` in \`openai-compatible-client.ts\` calls \`undiciFetch\` directly. **No queue, no depth cap.**

### 2. Circuit breakers

Old path wraps the fetch with \`ollamaBreakers.chat.execute(...)\` / \`openaiBreakers.chat.execute(...)\`. New path has no breaker. \`CircuitBreakerOpenError\` will never fire for chat post-migration.

## Proposed remediation

Wrap the new client functions — at the \`openai-compatible-client.ts\` layer, not at every call site — with:

1. A per-provider circuit breaker (replace the per-protocol \`ollamaBreakers\` / \`openaiBreakers\` with a \`providerId\`-keyed map).
2. A queue that respects \`LLM_CONCURRENCY\` / \`LLM_MAX_QUEUE_DEPTH\` for non-stream calls. Streaming remains bypassed, same as today's note in \`providerStreamChat\`.

This keeps consumers (routes + workers) ignorant of the queue/breaker layer, mirroring today's design.

## Also: file-path mismatch

Plan tasks 16 + 17 referenced \`llm-chat.ts\` and \`llm-ask.ts\` as separate files. In reality only \`llm-ask.ts\` exists; \`llm-chat.test.ts\` exercises the same routes via \`llmAskRoutes\`. Task 17 will do the actual body swap.

## Acceptance

- [ ] \`openai-compatible-client.ts\` \`chat\` / \`generateEmbedding\` pass through the \`enqueue\` queue.
[...]

Issue #257, opened 2026-04-20
follow-up(#256): wire BullMQ re-embed-all worker


laboef1900:
## Context

Issue #256 (multi-LLM-provider) introduces `POST /api/admin/embedding/reembed` that returns a \`jobId\` and handles the schema-change dimension path (truncate + ALTER TYPE + HNSW rebuild). What it does **not** yet do: enqueue and run an actual BullMQ job that iterates every page and calls \`embedText()\` per page.

## Deliverable

In \`backend/src/domains/llm/services/embedding-service.ts\`, replace the TODO in \`enqueueReembedAll\`:

\`\`\`ts
// TODO(#257): enqueue BullMQ job that iterates pages and calls embedText() per page.
\`\`\`

with a real \`queueService.enqueueJob('reembed-all', { jobId })\` call, backed by a worker that:
1. Selects all page IDs into a cursor.
2. For each page, computes \`embedText(body_text)\` via \`resolveUsecase('embedding')\`.
3. Inserts/updates the row in \`page_embeddings\`.
4. Emits progress (every 100 pages → job status) so the frontend banner can poll.

## Acceptance criteria

- [ ] \`POST /admin/embedding/reembed\` returns a \`jobId\` that progresses from 0 to \`pageCount\` over time.
- [ ] Job is idempotent — re-invoking while one is already running returns the existing \`jobId\` (the spec already promises this but the current stub does not enforce it).
- [ ] Playwright E2E \`e2e/llm-providers.spec.ts\` asserts the job completes within 60s for the test fixture.

## Depends on

- #256 merges first.

Issue #214, opened 2026-04-17
Per-use-case LLM provider/model configuration + move SUMMARY_MODEL to admin settings


laboef1900:
## Problem

Currently the shared LLM configuration is a single provider + model pair (`llm_provider`, `ollama_model`, `openai_model`) applied uniformly to every use case. Background workers for summary and quality analysis, chat, auto-tagging, embeddings, etc. all end up pointing at the same provider/model unless overridden via env vars.

Two concrete pain points:

1. **No per-use-case provider routing.** Operators may want a small, fast local model for summaries/quality (e.g. `qwen3:4b` on Ollama) while routing interactive chat/RAG to a larger hosted OpenAI-compatible model — or vice versa. Today this requires either juggling env vars per worker or running multiple backends.
2. **`SUMMARY_MODEL` still lives in env.** `backend/src/domains/knowledge/services/summary-worker.ts:33` reads `process.env.SUMMARY_MODEL ?? process.env.DEFAULT_LLM_MODEL`. This is inconsistent with how `llm_provider`, `ollama_model`, `openai_model`, `embedding_model`, and `fts_language` are already persisted in `admin_settings` and editable in the admin UI. Admins have to redeploy to change the summary model.

`QUALITY_MODEL` (`backend/src/domains/knowledge/services/quality-worker.ts:20`) has the same problem.

### 1. Introduce per-use-case LLM assignments in admin settings

Define a small fixed set of \"LLM use cases\" and let the admin pick `{provider, model}` per use case, falling back to the shared default when unset.

Suggested use cases:
- `chat` — user-facing chat / article generation / improvement
- `summary` — background summary worker
- `quality` — background quality-analysis worker
- `auto_tag` — auto-tagger
- `embedding` — already modeled separately, keep as-is

Data model: extend `admin_settings` with keys like `llm_usecase_summary_provider`, `llm_usecase_summary_model`, `llm_usecase_quality_provider`, `llm_usecase_quality_model`, etc. Keep the existing `llm_provider` / `ollama_model` / `openai_model` rows as the shared default that applies when a use-case override is empty.

Backend: add a resolver in `core/services/admin-settings-service.ts` (or a new `llm-assignments-service.ts`) that takes a use-case name and returns the resolved `{provider, model}`, merging use-case override → shared default → env fallback. Callers (`summary-worker`, `quality-worker`, `auto-tagger`, chat routes) ask the resolver instead of reading env directly.

[...]

Issue #12, opened 2026-03-30
Editor: Show resize cursor when hovering over table column borders


laboef1900:
## Summary

When hovering over a table column border in the article editor, the mouse cursor should change to a **resize indicator** (e.g., `col-resize`) to signal that the column can be resized by dragging.

## Current Behavior

The mouse cursor does not change when hovering over column borders in a table, giving no visual feedback that columns are resizable.

## Expected Behavior

- When hovering over a column border, the cursor should change to a resize cursor (`col-resize` / `↔`)
- The indicator should appear on the border between columns, making it clear where to click and drag
- Column resizing via drag should work smoothly with the visual indicator

## Acceptance Criteria

- [ ] Resize cursor (`col-resize`) shown when hovering over table column borders
- [ ] Visual feedback is clear and precise on the draggable border area
- [ ] Works consistently across all tables in the editor

laboef1900:
Implemented in PR #26: `column-resize-handle` with `cursor: col-resize` in index.css.

PR #259, opened 2026-04-21
feat: multi-LLM-provider configuration

### LINKED ISSUE #256: feat: multi-LLM-provider configuration (N named providers, per-use-case assignment incl. embedding)
## Summary

Replace today's two-slot LLM admin config (`{ollama, openai}` in `admin_settings`) with an N-provider table of `openai-compatible` endpoints that can be individually assigned to five use cases — **chat**, **summary**, **quality**, **auto_tag**, **embedding** — with a re-embed guardrail when the embedding provider or model changes.

## Motivation

Operators need to wire different LLM servers to different workloads (e.g. local GPU box for chat, cloud OpenAI for batch summary, a dedicated embedding endpoint) without contorting the two-slot model. Ollama already speaks the OpenAI-compatible `/v1` protocol, so the split between the two provider types is a historical artifact.

## Scope decisions (locked)

| # | Question | Decision |
|---|----------|----------|
| Q1 | Provider cardinality & types | N named providers, single type `openai-compatible` |
| Q2 | Embedding as a use case | Fifth use case with re-embed guardrail |
| Q3 | Admin-only vs per-user | Admin-only |
| Q4 | Shared-default / inheritance | `is_default` provider + per-provider `default_model`; use cases may inherit |
| Q5 | Delete semantics | Block deletion when referenced or when `is_default` |

## Deliverables

- Two new tables: `llm_providers`, `llm_usecase_assignments` (FK `ON DELETE RESTRICT`, partial unique index enforcing one default)
- Unified `openai-compatible-client` replacing `ollama-service` + `openai-service` + `ollama-provider`
- `resolveUsecase(usecase)` replacing `resolveUserProvider` and `providerStreamChat*` / `providerChatForUsecase`
- REST routes: `/api/admin/llm-providers` CRUD + `test` + `models` + `set-default`; `/api/admin/llm-usecases` GET/PUT; `/api/admin/embedding/reembed`
- Admin UI (Settings → LLM): provider list CRUD + use-case grid with embedding row + re-embed banner
- Migration 054 seeds new rows from legacy `admin_settings` keys, then deletes them
- Doc + ADR updates

## Design + plan

- Spec: [\`docs/superpowers/specs/2026-04-20-multi-llm-providers-design.md\`](docs/superpowers/specs/2026-04-20-multi-llm-providers-design.md)
- Implementation plan: [\`docs/superpowers/plans/2026-04-20-multi-llm-providers.md\`](docs/superpowers/plans/2026-04-20-multi-llm-providers.md) (36 TDD-ordered tasks across 8 phases)
- Branch: \`feature/multi-llm-providers\`

## Acceptance criteria

1. Admin can create ≥ 2 providers in the UI and assign any of them to any of the five use cases.
2. Changing an assignment takes effect on the next request without a restart.
3. Deleting a provider that is in use returns HTTP 409 with the list of blocking use cases.
4. Upgrade from a pre-054 install preserves the currently-active provider and model in the default + chat slots; no user action required.
5. Changing the embedding provider or model shows a confirmation dialog and surfaces a re-embed action that successfully re-embeds all pages.
6. \`npm test\`, \`npm run lint\`, \`npm run typecheck\` pass in both workspaces.

## Security notes

- API keys AES-256-GCM-encrypted via the existing \`encryptPat\`/\`decryptPat\` (same primitive Confluence PATs already use).
- GET responses return \`hasApiKey\` + \`keyPreview\` (last 4) only; plaintext keys never leave the backend.
- \`baseUrl\` validated through the existing \`ssrf-guard\` util.
- Admin-only routes, Zod validation at every boundary, rate-limited \`/test\` + \`/models\`.
- Every provider mutation emits an \`LlmAuditEntry\` via \`emitLlmAudit()\`.

## Out of scope (YAGNI)

- Per-user provider overrides
- Non-OpenAI-compatible protocols (Anthropic native, Vertex)
- Automated re-embed triggered by dimension-mismatch detection
- Model aliasing / cost tracking / routing

laboef1900:
Closes #256.

Follow-ups (tracked separately, not in this PR):
- #257 — BullMQ-backed worker that actually iterates pages + calls `embedText()` after the re-embed job is enqueued. This PR ships the schema-change path (TRUNCATE + ALTER TABLE + rebuild HNSW) and the `/admin/embedding/reembed` endpoint, but the worker loop is a follow-up.
- #258 — already merged at `22add6e` (queue + per-provider circuit breakers inside `openai-compatible-client.ts`).

---

# Multi-LLM-Provider Configuration — Design

**Date:** 2026-04-20
**Status:** Draft — pending implementation plan
**Scope:** `compendiq-ce` (backend + frontend + contracts)
**Supersedes:** Parts of issue #214 (per-use-case assignments); the two-slot `llm_provider` model in `admin_settings_service`.

---

## 1. Motivation

Today Compendiq stores exactly two LLM "slots" in `admin_settings`:

- **Ollama** — `OLLAMA_BASE_URL` env var + `ollama_model` setting.
- **OpenAI-compatible** — `openai_base_url` + `openai_api_key` + `openai_model` settings.

Per-use-case assignments (chat / summary / quality / auto_tag — issue #214) can
only pick between those two slots. Embedding is a single global `embedding_model`
piggy-backing on whichever slot is active.

Operators need to wire different LLM servers to different workloads — e.g. a
local GPU box for chat, a cloud OpenAI endpoint for batch summary work, a
different model entirely for embeddings — without contorting the two-slot model.
This spec defines the admin-managed multi-provider replacement.

## 2. Scope decisions (locked with the user)

| # | Question | Decision |
|---|----------|----------|
| Q1 | Provider cardinality & types | **N named providers, single type `openai-compatible`.** No separate `ollama` type — Ollama exposes `/v1/chat/completions`, `/v1/embeddings`, `/v1/models` and every other backend already speaks this protocol. |
| Q2 | Embedding as a use case | **Fifth use case with re-embed guardrail.** Changing provider/model shows a confirmation and surfaces a "Re-embed all pages" action afterwards. |
| Q3 | Admin-only vs per-user | **Admin-only.** Matches the current model; background workers have no user context; user-level overrides are an explicit future scope item. |
| Q4 | Shared-default / inheritance | **`is_default` provider + `default_model` per provider.** Use cases can inherit (both `NULL` in the DB) or override provider-only or provider+model. |
| Q5 | Delete semantics | **Block deletion when referenced or when `is_default`.** Admin must reassign use cases first. `ON DELETE RESTRICT` FK enforces this at the DB level. |

## 3. Data model

Two new tables replace the flat `admin_settings` keys for LLM configuration.

### 3.1 `llm_providers`

```sql
CREATE TABLE llm_providers (
  id             UUID         PRIMARY KEY DEFAULT gen_random_uuid(),
  name           TEXT         NOT NULL UNIQUE,
  base_url       TEXT         NOT NULL,                    -- write-path normalizes to end with /v1
  api_key        TEXT         NULL,                        -- AES-256-GCM via encryptPat()
  auth_type      TEXT         NOT NULL DEFAULT 'bearer',   -- 'bearer' | 'none'
  verify_ssl     BOOLEAN      NOT NULL DEFAULT TRUE,
  default_model  TEXT         NULL,
  is_default     BOOLEAN      NOT NULL DEFAULT FALSE,
  created_at     TIMESTAMPTZ  NOT NULL DEFAULT NOW(),
  updated_at     TIMESTAMPTZ  NOT NULL DEFAULT NOW()
);

-- At most one default row.
CREATE UNIQUE INDEX llm_providers_one_default
  ON llm_providers (is_default) WHERE is_default;
```

**Service-layer invariant:** the default flag is *swap-only* —
`setDefaultProvider(id)` runs `UPDATE SET is_default=FALSE WHERE is_default=TRUE`
and `UPDATE SET is_default=TRUE WHERE id=$1` in the same transaction. There is
**no** code path that clears the default without promoting a replacement. Combined
with `deleteProvider` blocking when `is_default=true`, this prevents a
zero-default terminal state where every `resolveUsecase` throws.

### 3.2 `llm_usecase_assignments`

```sql
CREATE TABLE llm_usecase_assignments (
  usecase      TEXT         PRIMARY KEY,                   -- 'chat' | 'summary' | 'quality' | 'auto_tag' | 'embedding'
  provider_id  UUID         NULL REFERENCES llm_providers(id) ON DELETE RESTRICT,
  model        TEXT         NULL,
  updated_at   TIMESTAMPTZ  NOT NULL DEFAULT NOW(),
  CHECK (usecase IN ('chat', 'summary', 'quality', 'auto_tag', 'embedding'))
);
```

**Resolution rules** (pure function, no cache):

| `provider_id` | `model` | Resolved provider | Resolved model |
|---------------|---------|-------------------|----------------|
| `NULL` | `NULL` | the `is_default` row | that row's `default_model` |
| set   | `NULL` | the referenced row    | that row's `default_model` |
| set   | set    | the referenced row    | the override string |
| `NULL` | set   | the `is_default` row | the override string (rare; legal) |

If no default exists (fresh install, nothing configured), the resolver returns
an error surfaced as a 503 by the API boundary so workers can back off.

### 3.3 Removed `admin_settings` keys

Deleted by migration `054` after seeding the new tables:

- `llm_provider`
- `ollama_model`
- `openai_base_url`
- `openai_api_key`
- `openai_model`
- `embedding_model`
- `llm_usecase_chat_provider`, `llm_usecase_chat_model`
- `llm_usecase_summary_provider`, `llm_usecase_summary_model`
- `llm_usecase_quality_provider`, `llm_usecase_quality_model`
- `llm_usecase_auto_tag_provider`, `llm_usecase_auto_tag_model`

Retained (non-LLM-provider): `embedding_dimensions`, `fts_language`, SMTP,
OIDC, license, AI safety flags, all others.

## 4. Backend API

New routes under `backend/src/routes/llm/`, split into two files:

### 4.1 `llm-providers.ts`

```
GET    /api/admin/llm-providers                  list  (api_key masked)
POST   /api/admin/llm-providers                  create
PATCH  /api/admin/llm-providers/:id              partial update (omit apiKey = keep)
DELETE /api/admin/llm-providers/:id              409 if referenced or is_default
POST   /api/admin/llm-providers/:id/set-default  atomic flip of is_default
POST   /api/admin/llm-providers/:id/test         { connected, error?, sampleModelsCount }
GET    /api/admin/llm-providers/:id/models       passthrough /v1/models -> [{ name }]
```

List/GET responses use shape:

```ts
interface LlmProviderDTO {
  id: string;
  name: string;
  baseUrl: string;
  authType: 'bearer' | 'none';
  verifySsl: boolean;
  defaultModel: string | null;
  isDefault: boolean;
  hasApiKey: boolean;
  keyPreview: string | null;   // e.g. "…ab12", never full key
  createdAt: string;
  updatedAt: string;
}
```

### 4.2 `llm-usecases.ts`

```
GET /api/admin/llm-usecases  -> {
  [usecase]: {
    providerId: string | null,   // raw override
    model: string | null,        // raw override
    resolved: {                  // what the resolver would return right now
      providerId: string,
      providerName: string,
      model: string,
    }
  }
}

PUT /api/admin/llm-usecases  -> batch upsert using same {undefined = leave,
                                null = clear, value = set} tri-state semantics
                                as today's upsertUsecaseLlmAssignments.
```

### 4.3 Embedding guardrail

```
POST /api/admin/embedding/reembed  -> { jobId: string, pageCount: number }
```

Enqueues a full re-embed through the existing BullMQ embedding worker. Called
by the frontend after the user confirms the re-embed dialog. Idempotent:
re-invoking while a job is pending returns the existing `jobId`.

### 4.4 Contracts

`packages/contracts/src/llm.ts` adds:

- `LlmProviderSchema`, `LlmProviderInputSchema`, `LlmProviderUpdateSchema`
- `UsecaseAssignmentSchema`, `UsecaseAssignmentsSchema`, `UpdateUsecaseAssignmentsSchema`
  (rewritten to key off `providerId` instead of the old `'ollama' | 'openai'` union)
- `LlmUsecase = 'chat' | 'summary' | 'quality' | 'auto_tag' | 'embedding'`

### 4.5 Auth, validation, rate limiting

- Every `/api/admin/llm-*` route: `fastify.authenticate` + admin-role gate.
- Zod validation on path + body + query using `@compendiq/contracts` schemas.
- `/test` and `/models` routes rate-limited via the existing Fastify rate-limit
  plugin (same bucket as other admin discovery endpoints).
- `baseUrl` validated through `core/utils/ssrf-guard.ts`; rejects loopback +
  RFC1918 unless `LLM_ALLOW_PRIVATE_URLS=true` (already-in-use escape hatch).
  Before persisting, `baseUrl` is trimmed, has trailing slashes stripped, and
  `/v1` appended if missing — matching today's `getConfig` behavior in
  `openai-service.ts`.

### 5.1 Collapse the provider-specific services

Remove:

- `backend/src/domains/llm/services/ollama-service.ts`
- `backend/src/domains/llm/services/ollama-provider.ts`
- `backend/src/domains/llm/services/openai-service.ts`

Replace with `backend/src/domains/llm/services/openai-compatible-client.ts`
exporting pure functions that take a config arg — no module-level caches of
URL / key:

```ts
export interface ProviderConfig {
  providerId: string;
  baseUrl: string;
  apiKey: string | null;
  authType: 'bearer' | 'none';
  verifySsl: boolean;
}

export async function* streamChat(cfg: ProviderConfig, model: string,
                                  messages: ChatMessage[], signal?: AbortSignal)
                                  : AsyncGenerator<StreamChunk>;
export async function chat(cfg: ProviderConfig, model: string,
                           messages: ChatMessage[]): Promise<string>;
export async function generateEmbedding(cfg: ProviderConfig, model: string,
                                        text: string | string[]): Promise<number[][]>;
export async function listModels(cfg: ProviderConfig): Promise<LlmModel[]>;
export async function checkHealth(cfg: ProviderConfig): Promise<HealthResult>;
```

A per-`providerId` `undici.Agent` is memoized in-module so TLS / keep-alive
isn't rebuilt per request. The cache is invalidated when a provider row is
updated or deleted (see §5.3).

### 5.2 Resolver

`backend/src/domains/llm/services/llm-provider-resolver.ts` exposes:

```ts
export async function getProvider(providerId: string): Promise<ProviderConfig>;
export async function resolveUsecase(usecase: LlmUsecase)
    : Promise<{ config: ProviderConfig; model: string }>;
export async function listProviders(): Promise<ProviderConfig[]>;
```

`resolveUsecase` runs a single `LEFT JOIN` against `llm_usecase_assignments` +
`llm_providers` + the default row, applies the §3.2 resolution rules in
memory, and returns the resolved config + model.

**No caching on the assignment row.** Every `resolveUsecase` call hits PG
(one indexed query). This preserves the "changes take effect without restart"
acceptance criterion that was explicit for issue #214 and carries forward.

**Provider config *is* cached** by `providerId` in memory. Any `PATCH`,
`DELETE`, or `set-default` write bumps a module-scoped `version` counter
and invalidates the cache.

**Cache scope:** the version counter is **process-local**. In a horizontally
scaled deployment (multiple Node workers behind a load balancer), a write
routed to worker A does not invalidate worker B's cache until worker B makes
its next `PATCH` / `DELETE` / `set-default` call of its own, or restarts. CE
currently runs single-process; this is acceptable for the target deployment
scale. If we scale horizontally later, switch to a Postgres `LISTEN/NOTIFY`
fan-out on an `llm_providers_changed` channel. Recorded as a known constraint
rather than a bug.

**In-flight reads:** a `PATCH` that arrives while `resolveUsecase('chat')` is
already holding a `cfg` object leaves that in-flight call on the stale config.
This is intended — an in-flight HTTP call cannot be retargeted mid-flight, and
the next call after the write picks up the new config.

### 5.3 Call-site migration

| Today | Replaced by |
|-------|-------------|
| `resolveUserProvider(userId)` | removed |
| `providerStreamChat(userId, ...)` | `resolveUsecase('chat')` + `streamChat(cfg, model, …)` |
| `providerStreamChatForUsecase('summary', ...)` | `resolveUsecase('summary')` + `streamChat(…)` |
| `providerChatForUsecase('auto_tag', ...)` | `resolveUsecase('auto_tag')` + `chat(…)` |
| `ollamaGenerateEmbedding(text)` / `openai.generateEmbedding(text)` | `resolveUsecase('embedding')` + `generateEmbedding(…)` |

Consumers to update: `llm-chat`, `llm-ask`, `llm-summarize`, `llm-quality`,
`llm-generate`, `llm-improve`, `llm-diagram`, `generate-with-pdf`,
`analyze-quality`, `apply-improvement`, `improve-instruction`,
`improve-page-id`, `summary-worker`, `quality-worker`, `auto-tagger`,
`embedding-service`, `rag-service`.

### 5.4 Env-var behavior

| Env var | Fate |
|---------|------|
| `OLLAMA_BASE_URL`, `OPENAI_BASE_URL`, `OPENAI_API_KEY`, `LLM_BEARER_TOKEN` | Consulted **once** by the TS bootstrap in §6 to seed initial rows on a fresh install. Runtime reads are removed. A warning is logged if they are set on a later boot while `llm_providers` is already non-empty. |
| `LLM_VERIFY_SSL` | Default for the `verify_ssl` column on newly-created rows; the row value always wins at runtime. |
| `LLM_STREAM_TIMEOUT_MS`, `LLM_CONCURRENCY`, `LLM_MAX_QUEUE_DEPTH`, `LLM_CACHE_TTL`, `LLM_AUTH_TYPE` | Retained as operational knobs (not per-provider). `LLM_AUTH_TYPE` becomes a default for new rows only. |
| `EMBEDDING_MODEL`, `EMBEDDING_DIMENSIONS` | Deprecated. `EMBEDDING_MODEL` is consulted once during migration seed. `EMBEDDING_DIMENSIONS` is inferred from the embedding use case's model at re-embed time (carried by the existing `admin_settings['embedding_dimensions']` row, which is retained). |
| `SUMMARY_MODEL`, `QUALITY_MODEL`, `DEFAULT_LLM_MODEL` | Consulted once during migration seed; warning logged if set afterwards. |

## 6. Frontend UI

Settings → LLM tab (`frontend/src/features/settings/panels/LlmTab.tsx`)
is reorganized into two stacked sections.

### 6.1 Providers section

```
┌────────────────────────────────────────────────────────────────┐
│ Providers                                           [+ Add]    │
├────────────────────────────────────────────────────────────────┤
│ ● GPU Box (Ollama)      http://gpu:11434/v1    [default]  ⋯   │
│ ● OpenAI Prod           https://api.openai.com/v1          ⋯   │
│ ● LM Studio (dev)       http://192.168.1.5:1234/v1         ⋯   │
└────────────────────────────────────────────────────────────────┘
```

- Status dot: green / red from the last `/test` call, grey when never tested.
- `[default]` badge on the `is_default` row.
- `⋯` menu: Edit · Set as default · Test · Delete. Delete is disabled with a
  tooltip showing the blocking use cases ("Referenced by: chat, summary").
- `[+ Add]` opens a modal with fields `name`, `baseUrl`, `apiKey` (masked),
  `authType` radio (Bearer / None), `verifySsl` toggle, `defaultModel`
  (dropdown; populated after a successful in-modal `Test`).
- Edit modal shows `keyPreview` + a "Replace key" action; the stored key is
  left alone unless explicitly replaced. Base URL changes are allowed; the
  resolver picks them up on the next call.

### 6.2 Use-case assignments section

```
Use case       │ Provider                     │ Model                   │ → Resolves to
───────────────┼──────────────────────────────┼─────────────────────────┼───────────────────────────
Chat           │ [Inherit default       ▼]    │ [Inherit provider's ▼]  │ GPU Box / qwen3:4b
Summary worker │ [OpenAI Prod           ▼]    │ [gpt-4o-mini        ▼]  │ OpenAI Prod / gpt-4o-mini
Quality worker │ [Inherit default       ▼]    │ —                       │ GPU Box / qwen3:4b
Auto-tag       │ [GPU Box (Ollama)      ▼]    │ [Inherit provider's ▼]  │ GPU Box / qwen3:4b
Embedding  ⚠   │ [GPU Box (Ollama)      ▼]    │ [bge-m3             ▼]  │ GPU Box / bge-m3
```

- Provider column lists all providers + "Inherit default".
- Model column lazily fetches that provider's `/v1/models` via `useQuery`
  keyed by `providerId`; no fetch for "Inherit" rows.
- The **Embedding row** is a three-step flow, not a one-click save:
  1. **Probe.** On save, the backend calls `/v1/embeddings` on the chosen
     provider with a one-token test input and reads back the vector length `N`.
     Failure → the save is rejected with the upstream error; nothing persists.
  2. **Dimension-change branch.** If `N !== current_embedding_dimensions`
     (stored in `admin_settings['embedding_dimensions']`, today `1024`), the
     UI shows a **second** confirmation with the new dimension and the heavier
     warning *"This will delete all existing embeddings (N pages) and rewrite
     the `page_embeddings.embedding` column type — the HNSW index will be
     rebuilt. Continue?"* On confirm, the reembed job runs: `TRUNCATE
     page_embeddings` → `ALTER TABLE page_embeddings ALTER COLUMN embedding
     TYPE vector(N)` → update `admin_settings['embedding_dimensions']` → drop
     and recreate the HNSW index → enqueue re-embed for every page. Dimension
     mismatch is the headline risk pgvector imposes; the plan must treat it
     as a schema change, not just a data rewrite.
  3. **Same-dimension branch.** If `N === current`, we only enqueue re-embed
     with a single-confirmation dialog: *"This will make all existing vectors
     (N pages) inconsistent with the new model. You will need to re-embed.
     Continue?"*

  After either confirm path, a banner at the section top shows the re-embed
  job progress via the existing embedding job status endpoint.

### 6.3 Files touched / added

| Path | Action |
|------|--------|
| `frontend/src/features/settings/panels/LlmTab.tsx` | Rewrite top half, keep the use-case grid shape, reshape data sources. |
| `frontend/src/features/settings/panels/ProviderListSection.tsx` | New. |
| `frontend/src/features/settings/panels/ProviderEditModal.tsx` | New. |
| `frontend/src/features/settings/panels/UsecaseAssignmentsSection.tsx` | Extract existing inner component; generalize to N providers. |
| `frontend/src/features/settings/panels/EmbeddingReembedBanner.tsx` | New. |
| `packages/contracts/src/llm.ts` | Add new schemas, rewrite `UsecaseAssignments` against `providerId`. |

Status colors honor ADR-010 (green / red / yellow / blue / purple / gray).
Framer Motion entrance animations reuse the staggered pattern from neighboring
panels. Backdrop-blur cards via the `glass-card` utility.

### 7.1 Migration `054_llm_providers.sql`

Single file, idempotent, runs inside a transaction:

1. `CREATE TABLE llm_providers (…)` and `llm_usecase_assignments (…)` as §3.
2. `CREATE UNIQUE INDEX llm_providers_one_default …`.
3. **Seed provider rows** from existing `admin_settings` via an anonymous
   `DO $$ … $$` block. Migrations cannot read env vars — env-only installs
   are handled by the TS bootstrap in §7.2.
   - If `openai_base_url` OR `openai_api_key` OR `openai_model` is present →
     insert a row named `"OpenAI"` with those values. `api_key` is copied
     as-is (already AES-encrypted). `default_model` ← `openai_model`.
     `base_url` normalized to end with `/v1`; if `openai_base_url` is absent
     use `https://api.openai.com/v1`.
   - If `ollama_model` is present OR `llm_provider = 'ollama'` →
     insert a row named `"Ollama"` with a sentinel `base_url =
     'http://localhost:11434/v1'` (the migration cannot read
     `OLLAMA_BASE_URL`). The TS bootstrap rewrites the sentinel on first
     boot. `default_model` ← `ollama_model`.
   - Set `is_default = true` on whichever row matches the old `llm_provider`
     value (`'ollama'` → Ollama row, `'openai'` → OpenAI row). If neither
     provider row was inserted above (totally fresh install with no LLM
     admin_settings at all), `is_default` is deferred to the TS bootstrap.
4. **Seed `llm_usecase_assignments`** from the eight old `llm_usecase_*_*`
   keys. Look up the new `provider_id` by name (`"Ollama"` / `"OpenAI"`).
   Missing keys → row omitted (inherits at runtime).
5. **Seed embedding row**: `{usecase: 'embedding', provider_id: <is_default id>,
   model: admin_settings['embedding_model']}`.
6. **Delete the migrated `admin_settings` keys** listed in §3.3 in the same
   transaction so there is no dual source of truth post-migration.

### 7.2 TS bootstrap

Runs once on startup, after migrations, guarded by a `seededFromEnv`-style Set
so it is safe to rerun:

- If `llm_providers` is empty (fresh install) and `OLLAMA_BASE_URL` or
  `OPENAI_BASE_URL` is set → insert the corresponding row(s) and mark the
  first as default.
- If an `"Ollama"` row has the sentinel `http://localhost:11434/v1` AND
  `OLLAMA_BASE_URL` differs → replace its `base_url`.
- If no `is_default` row exists after the above → promote the oldest row
  to default.
- Log a WARN if any of `DEFAULT_LLM_MODEL`, `SUMMARY_MODEL`, `QUALITY_MODEL`,
  `OPENAI_BASE_URL`, `OPENAI_API_KEY`, `LLM_BEARER_TOKEN`, `EMBEDDING_MODEL`
  are set after migration — they no longer have effect.

### 7.3 Rollback plan

Additive schema with destructive key removal. No down-migration provided.
Rollback = restore from the PG backup taken immediately before the `054`
migration. The release runbook paragraph is added to
`docs/ARCHITECTURE-DECISIONS.md`.

### 7.4 Doc & ADR updates in the same PR

- New ADR recording Q1–Q5 decisions (`docs/ARCHITECTURE-DECISIONS.md`
  appendix; next available number).
- `docs/architecture/03-backend-domains.md` — update the LLM domain box to
  reference `llm-provider-resolver` instead of the two service files.
- `docs/architecture/06-data-model.md` — add `llm_providers` +
  `llm_usecase_assignments` tables + the FK to `llm_providers.id`.
- `CLAUDE.md` + `AGENTS.md` — rewrite the "External Services" row for LLM
  (delete per-slot env vars) and trim the "Environment" section accordingly.

### 8.1 Backend

- **Unit — resolver** (`llm-provider-resolver.test.ts`). Real PG via
  `test-db-helper.ts`. Covers every branch in §3.2 plus: missing-default,
  provider-deleted-mid-call (expect `RESTRICT` to have blocked it earlier),
  cache invalidation after PATCH.
- **Route — providers** (`llm-providers.test.ts`). CRUD happy paths, 409 on
  delete-while-referenced, 409 on delete-while-default, atomic `set-default`
  under two concurrent callers, `/test` with a fake `/v1/models` backend
  (mock at the undici boundary, not the service function).
- **Route — use-cases** (`llm-usecases.test.ts`). Batch upsert tri-state,
  resolved block accuracy across all five use cases.
- **Service — client** (`openai-compatible-client.test.ts`). Single suite for
  chat / stream / embed / models / health against a minimal fake `/v1` server.
  Existing `ollama-service.test.ts` + `openai-service.test.ts` are removed.
- **Migration** (`__tests__/054_llm_providers.test.ts`). Seeds pre-054
  `admin_settings` state, runs the migration, asserts new rows + absence of
  old keys.

### 8.2 Frontend

- `LlmTab.test.tsx` — rewritten.
- `ProviderListSection.test.tsx`, `ProviderEditModal.test.tsx`,
  `EmbeddingReembedBanner.test.tsx` — new. MSW-style `fetch` mocks
  (`vi.spyOn(globalThis, 'fetch')`); no component mocking.

### 8.3 E2E

One new Playwright spec (`e2e/llm-providers.spec.ts`): admin adds a provider,
sets it as default, assigns it to chat, issues a chat request, asserts the
response.

### 8.4 Coverage

80 %+ line coverage per workspace policy. 100 % on the §3.2 resolution rules.

## 9. Security

- API keys encrypted at rest via existing `encryptPat` / `decryptPat`
  (AES-256-GCM with `PAT_ENCRYPTION_KEY`). Same primitive that already
  protects Confluence PATs — no new crypto surface.
- GET responses never include plaintext; only `hasApiKey` + `keyPreview`
  (last 4 chars). `PATCH` accepts `apiKey` to replace; omitting it keeps the
  stored value untouched.
- `baseUrl` validated via existing `ssrf-guard` utility. Rejects loopback +
  RFC1918 unless `LLM_ALLOW_PRIVATE_URLS=true` (existing escape hatch for the
  localhost Ollama case). Applies to create + update.
- Zod validation on every route boundary using `@compendiq/contracts`
  schemas. Parameterized SQL only via `pg` driver.
- `fastify.authenticate` + admin-role check on every `/api/admin/llm-*`
  route.
- Rate limits on `/test` and `/models` (reuse existing Fastify plugin
  config) so the routes can't be used as an egress probe.
- Audit log: every provider create / update / delete / set-default emits an
  `LlmAuditEntry` via `emitLlmAudit()` so EE audit pipelines pick it up.

## 10. Out of scope (YAGNI)

- Per-user provider overrides.
- Non-OpenAI-compatible protocols (native Anthropic, Vertex).
- Automated re-embed triggered by dimension mismatch detection.
- Model aliasing / cost tracking / routing.
- Frontend model auto-discovery on startup.

## 11. Acceptance criteria

1. Admin can create ≥ 2 providers of type openai-compatible in the UI and
   assign any of them to any of the five use cases (chat, summary, quality,
   auto_tag, embedding).
2. Changing an assignment takes effect on the next request without a restart.
3. Deleting a provider that is in use returns HTTP 409 with the list of
   blocking use cases.
4. Upgrade from a pre-054 install preserves the currently-active provider and
   model in the default + chat slots; no user action required for chat to keep
   working.
5. Changing the embedding provider or model shows a confirmation dialog and
   surfaces a re-embed action that successfully re-embeds all pages.
6. `npm test`, `npm run lint`, `npm run typecheck` pass in both workspaces.

---

**Next step:** transition to the `writing-plans` skill to produce a
step-by-step implementation plan with surgical edits, test ordering, and
rollback.


laboef1900 (commented):
## Automated PR Review

**PR**: #259 — feat: multi-LLM-provider configuration
**Branch**: `feature/multi-llm-providers` -> `dev`
**Diff size**: 17,434 lines (+8,715 / -6,121) across 100 files, 40 commits
**Reviewed at**: 2026-04-21 09:00 UTC

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | clean across all 3 workspaces |
| Lint | PASS | clean (one `eslint-plugin-boundaries` v5→v6 migration notice, not a lint failure) |
| Backend Tests | PASS | 131 files, 1893 tests, all green (real PG on :5433) |
| Frontend Tests | PASS | 149 files, 1805 tests, all green |
| Production Build | PASS | vite build succeeded in 10s |
| npm audit | PASS | 0 production vulnerabilities |
| Branch Target | PASS | feature/* -> dev (correct) |

### Docker App Test

| Check | Status |
|-------|--------|
| Containers Healthy | PASS (backend, frontend, postgres, redis all healthy) |
| Login | PASS (registered first admin via `/api/auth/register`, promoted to admin, token issued) |
| Backend `/api/health` | PASS (returns full health incl. queue metrics and legacy breaker fields) |
| Backend `/api/health/ready` | PASS |
| Frontend loads | PASS (HTTP 200) |
| Migration 054 | PASS (applied in 2ms, no errors in logs) |
| LLM provider CRUD | PASS (live-verified create, list, set-default) |
| Use-case resolution | PASS (live-verified inherit-default, explicit override, model override) |
| No error logs | PASS (no fatal/unhandled/crash signals beyond intentional test-path errors) |

Note: Port 3052 was in use by an existing `compendiq` stack, so the reviewer spun up on ports 3055/8085/5434/6380 via overrides. Volumes were isolated via a compose overlay so the reviewer DB is disposable.

### E2E Browser Tests

| Project | Status |
|---------|--------|
| Playwright (`e2e/llm-providers.spec.ts`) | SKIPPED (did not run browser tests; the spec self-skips when LLM endpoint isn't reachable — the stack's configured Ollama URL `localhost:11434/v1` is not reachable from inside the reviewer container) |

### Code Review Findings

| 1 | WARNING | `backend/src/domains/llm/services/llm-queue.ts:103-123` | `LLM_CONCURRENCY` env var is documented in `.env.example`, `CLAUDE.md` (this PR's own edit), **and the PR's own design spec §5.4** as a retained operational knob, but it is NOT read anywhere. `initLlmQueue()` only consults the `admin_settings` row `llm_concurrency`. `setConcurrency()` is exported but not called from any env-reading path. The env var is dead code. Either wire it (fallback to env when the DB row is absent) or remove from docs and spec. Same applies to `LLM_MAX_QUEUE_DEPTH` and `LLM_STREAM_TIMEOUT_MS` — `LLM_STREAM_TIMEOUT_MS` is read as `DEFAULT_TIMEOUT_MS` at module import (line 20) only, which means a later DB setting can override it but it can't be changed after startup. Documented behaviour says they're "retained as operational knobs"; actual behaviour is DB-driven. |
| 2 | WARNING | `backend/src/routes/foundation/health.ts:4, 149-150` and `backend/src/core/services/circuit-breaker.ts:159-194` | `/api/health` still exports `circuitBreakers.ollama` and `circuitBreakers.openai` status fields, but nothing in the new code path ever increments `ollamaBreakers.*` / `openaiBreakers.*`. The client now goes through `getProviderBreaker(cfg.providerId)`, which is a separate `Map<string, CircuitBreaker>`. Result: `/api/health` always reports legacy breakers CLOSED even when every per-provider breaker is OPEN. I verified this live — the running stack's health payload shows `circuitBreakers.ollama.chat.state=CLOSED` unconditionally. Operators relying on this signal will be misled. Either remove the legacy breaker block from the response or replace it with an aggregate over `providerBreakers`. |
| 3 | WARNING | `backend/src/domains/llm/services/embedding-service.ts:126-135` | `getEmbedBreakerNextRetryTime()` reads `ollamaBreakers.embed` and `openaiBreakers.embed`, which (see #2) are never incremented anymore. The bulk-embed loop at line 531 uses the return value to decide how long to back off when a circuit-breaker error is caught. Because the lookup always returns `null`, the wait loop falls back to the minimum buffer (`CIRCUIT_BREAKER_WAIT_BUFFER_MS`) regardless of when the breaker would actually re-close. Functional but suboptimal — re-embeds under provider outage will retry tighter than designed. Read from `getProviderBreaker(cfg.providerId).getStatus().nextRetryTime` instead, threaded through from the resolved embedding config. |
| 4 | WARNING | `backend/src/routes/llm/llm-providers.ts:47, 69` | SSRF guard rejections surface as HTTP 500 with "Internal Server Error" instead of 4xx. Live-verified: POSTing `{baseUrl: "http://127.0.0.1:9999/v1"}` or RFC1918 addresses throws `SsrfError` from `assertNonSsrfUrl`, and the default Fastify error handler returns 500. Admin users get no feedback explaining which part of the URL is disallowed. Wrap `assertNonSsrfUrl(input.baseUrl)` in a try/catch and `reply.code(400).send({error: 'Invalid baseUrl', reason: err.message})`, or add a typed handler for `SsrfError` in the app error pipeline. Same issue in the PATCH branch. |
| 5 | WARNING | `backend/src/routes/llm/llm-providers.ts:91-96, 111-116` | Exception-to-HTTP mapping uses message regex (`/default\|referenced/i`) and a blanket catch for set-default. Two brittleness points: (a) a concurrent `PUT /api/admin/llm-usecases` inserting a reference to a provider that `deleteProvider` is in the middle of deleting surfaces PG error code `23503` (FK violation) — the message doesn't contain "default" or "referenced", so the client gets a 500 instead of 409. (b) `set-default` catches everything and returns 404, swallowing the real cause. Prefer asserting error class / PG error code, or running the `deleteProvider` check + DELETE inside a `SERIALIZABLE` transaction with a retry. |
| 6 | WARNING | `backend/src/domains/llm/services/embedding-service.ts:906-945` | `enqueueReembedAll` does the schema-change half (TRUNCATE, ALTER TABLE, update `embedding_dimensions` setting, rebuild HNSW) but returns a synthetic `jobId = 'reembed-' + Date.now()` without enqueueing the actual per-page re-embed work. This means acceptance criterion #5 — "surfaces a re-embed action that successfully re-embeds all pages" — is only half-met after this PR merges: existing vectors are deleted, the column is resized, the index rebuilt, and then pages stay unembedded until a user manually hits the embed paths again. The PR description openly acknowledges this as follow-up #257, but operators need to be aware because a naive "I changed the embedding dimension" workflow will destroy retrieval until #257 ships. At minimum, the API should return `enqueued: 0` and log a WARN; ideally the UI should surface "re-embed worker not yet running" on the reembed banner. |
| 7 | INFO | `backend/src/routes/foundation/health.ts:41` | `checkLlm` hardcodes `providerId: 'health-probe'` when constructing the config for `providerCheckHealth(cfg)`. That means every health probe shares a single circuit breaker named `llm-provider:health-probe` across ALL providers the health endpoint tests. If the default provider is briefly unreachable, the breaker trips after 3 probes and stays OPEN for 30s — subsequent probes return "connected: false" without attempting the upstream, even if it recovers in that window. Minor observability quirk; pass the real `p.id` (already available in the row) to avoid this. |
| 8 | INFO | `backend/src/core/db/migrations/054_llm_providers.sql:30-118` | The seeding `DO $$ … $$` block is idempotent at the row level (`ON CONFLICT DO NOTHING` everywhere), but the `RETURNING id INTO openai_id/ollama_id` pattern only fires on actual INSERT. On a re-run where the rows already exist, `openai_id` / `ollama_id` stay NULL, so the subsequent `UPDATE … SET is_default=TRUE WHERE id = $var` becomes a no-op. That's safe (the is_default flag won't be clobbered if set manually) but means a migration that failed midway cannot self-heal the is_default state without operator intervention. If you want strict idempotency, add `IF openai_id IS NULL THEN SELECT id INTO openai_id FROM llm_providers WHERE name = 'OpenAI'; END IF;` after the INSERT. Not needed for the common case. |
| 9 | INFO | `backend/src/routes/llm/llm-providers.ts:160-171` | `/admin/llm-providers/:id/models` bubbles upstream fetch errors as HTTP 500 (live-verified against the seeded TestProvider that points at an unreachable baseUrl). Sibling `/test` correctly returns `{connected:false, error}` with HTTP 200 — this inconsistency is worth smoothing. Wrap the `clientListModels` call the same way. |
| 10 | INFO | `packages/contracts/src/llm.ts:14` | `LlmProviderInputSchema.apiKey` is `z.string().min(1).optional()`. Clients sending `{"apiKey": null}` get a 400 "expected string". The spec §4.1 talks about tri-state semantics only for PATCH; creating with an explicit `null` is a natural client gesture that's rejected here. Consider `.nullable().optional()` to accept both `undefined` and `null` as "no key". |
| 11 | INFO | `CLAUDE.md` (not updated beyond the External-Services row) | Spec §7.4 calls for updates to "CLAUDE.md + AGENTS.md — rewrite the 'External Services' row for LLM (delete per-slot env vars) **and trim the 'Environment' section accordingly.**" The External-Services row is trimmed, but (a) AGENTS.md wasn't touched at all in this PR (b) the Environment section in CLAUDE.md still lists `LLM_PROVIDER`, `OPENAI_BASE_URL`, `OPENAI_API_KEY` as current knobs, contradicting the new reality that they're "deprecated bootstrap fallbacks only". The deprecated markers *were* added for `DEFAULT_LLM_MODEL`, `SUMMARY_MODEL`, `QUALITY_MODEL` — apply the same treatment to the OpenAI/provider vars. |
| 12 | INFO | `frontend/src/features/settings/panels/ProviderEditModal.tsx:57-135` | Modal uses `role="dialog"` but lacks `aria-modal`, `aria-labelledby`, focus trap, and Esc-to-close. Headless UI / Radix Dialog is already a dep (spec mentions Radix primitives); consider using `@radix-ui/react-dialog` for accessibility. |

### Acceptance Criteria (Issue #256)

All 5 numbered criteria from the issue, verified where possible against the running stack:

| AC-1 | Admin can create ≥2 providers and assign any to any of the 5 use cases | PASS | Live: created `TestProvider` via POST, assigned to `chat` via PUT /admin/llm-usecases, resolved block showed correct provider |
| AC-2 | Changing an assignment takes effect on the next request without restart | PASS | Live: after PUT to chat assignment, immediate GET showed resolved provider switched — no restart. Verified mechanism: `bumpProviderCacheVersion()` called in `llm-usecases.ts:90` and in every provider mutation in `llm-provider-service.ts`; resolver re-queries assignment every call. |
| AC-3 | Deleting a provider in use returns HTTP 409 with blocking use-cases | PASS | Live: DELETE on referenced provider returned HTTP 409 `{"error":"Provider is referenced by: chat"}`. DELETE on default returned HTTP 409 `{"error":"Cannot delete the default provider..."}`. |
| AC-4 | Pre-054 upgrade preserves the active provider+model in default+chat | PASS (by migration test) | Did not have a pre-054 DB to probe live. Covered by `backend/src/core/db/migrations/__tests__/054_llm_providers.test.ts` which is green in the test run. Migration logic in the `DO $$` block at `054_llm_providers.sql:30-118` matches the spec exactly: seeds OpenAI/Ollama rows from `admin_settings`, copies encrypted `api_key` as-is, sets is_default from legacy `llm_provider` key, seeds per-use-case rows, seeds embedding row, deletes the migrated keys. |
| AC-5 | Changing embedding provider/model confirms, then successfully re-embeds all pages | PARTIALLY MET | Probe works (`POST /admin/embedding/probe` live — returned `fetch failed` against unreachable upstream, graceful). Dimension-change schema path works (ALTER TABLE + HNSW rebuild inside a transaction, dimension validated to 1–8192 int). BUT: the actual page-iteration + `embedText()` worker is stubbed (finding #6); `enqueueReembedAll` returns a synthetic jobId. PR body explicitly scopes this to follow-up #257. |
| AC-6 | `npm test`, `npm run lint`, `npm run typecheck` pass in both workspaces | PASS | Confirmed — see Static Checks section above. |

Additional spec validations:
- Security §9: API keys AES-256-GCM via `encryptPat`/`decryptPat` (reused, no new crypto surface) ✓. GET masks keys to `hasApiKey` + `keyPreview` (live: saw `"keyPreview":"…1234"`, plaintext never returned). ✓. `baseUrl` through `ssrf-guard` ✓ (live-verified loopback / RFC1918 / link-local all rejected at runtime — though with bad HTTP status; see #4). `fastify.authenticate` + `fastify.requireAdmin` on every route ✓. Rate limits on `/test` + `/models` ✓ (same admin bucket as spec §4.5). `emitLlmAudit` on every mutation ✓.
- Tests §8: Health.test.ts / setup.test.ts rewrites (user-flagged "verify not tautological"): legit — 13/19 `it()` blocks with real assertions about status codes, service-availability branching, and mock-call verification (e.g. `expect(mockCheckHealth).toHaveBeenCalledWith(expect.objectContaining({baseUrl: 'https://api.openai.com/v1'}))`). No `toBe(true).toBe(true)` patterns. `openai-compatible-client.test.ts:190-226` specifically verifies "opens after 3 failures and short-circuits the 4th call" by counting server hits — directly exercises the breaker claim. ✓.
- Legacy `ollama-service.ts`, `ollama-provider.ts`, `openai-service.ts`, `llm-provider.ts` all removed ✓ (only a comment mentions them in `prompts.ts`).
- Docs updates: ADR added (32 lines), `03-backend-domains.md` + `06-data-model.md` updated ✓ per CLAUDE.md Rule #6. CLAUDE.md partially updated (see #11). AGENTS.md not updated (see #11).

### Story Validation (Issue #256)

**Issue intent**: Replace the two-slot `{ollama, openai}` admin config with N-provider config, each individually assignable to chat/summary/quality/auto_tag/embedding, with a guardrail for embedding-dimension changes. Target: operators who want different LLMs per workload (local GPU for chat, cloud for batch summary, dedicated embedding endpoint).

**Story completion**: PARTIALLY ADDRESSED. AC-1/2/3/4/6 fully met with live evidence. AC-5 is half-met: the probe + schema-change flow works end-to-end (TRUNCATE, ALTER TABLE, HNSW rebuild all in one transaction), but the "then successfully re-embeds all pages" half is stubbed out. Author explicitly acknowledges this as #257 in the PR body and it is NOT represented as done.

**User journey**:
- Before: admin fiddles with 2 slots in `admin_settings`, embedding piggybacks on whichever slot is active, can't have 3 endpoints.
- After: admin visits Settings → LLM, adds an N-th provider, assigns it to any use-case, kill-switch for default is blocked, dimension change triggers a confirmation + schema rewrite. For everything except "the worker actually re-embeds pages after a dimension swap", the UX is delivered.

**Gaps**:
1. Re-embed worker is stubbed (see #6 and AC-5) — operators who flip the embedding provider and confirm the destructive dimension-change dialog will lose their existing embeddings AND not get new ones until #257 lands. This is the single most important caveat on this PR.
2. `LLM_CONCURRENCY` / `LLM_MAX_QUEUE_DEPTH` docs-vs-code mismatch (see #1) — spec promises env-var knobs that don't exist.
3. Observability regressions around the legacy breaker exports (see #2, #3) — `/api/health` readings and the embed retry loop both key off dead state.

**Evidence**: Live end-to-end verified via `POST /api/admin/llm-providers` (create), `GET /api/admin/llm-providers` (list masked), `POST /:id/set-default`, `PUT /api/admin/llm-usecases` (tri-state upsert), `GET /api/admin/llm-usecases` (resolved block), `DELETE /:id` → 409 for both default and referenced, migration 054 ran cleanly, SSRF guard blocks loopback/RFC1918/link-local, audit events emitted. Resolver CTE reviewed by hand for each of the four §3.2 resolution-rule cases — no SQL correctness bugs under concurrent writes (partial unique index on `is_default` prevents split-brain; `set-default` is transactional; `LIMIT 1` after `UNION ALL` is safe because the guard clauses ensure at most one branch matches).

### Suggested Actions

1. **[SHOULD FIX]** Either wire `LLM_CONCURRENCY` / `LLM_MAX_QUEUE_DEPTH` env vars into `initLlmQueue()` (fall back to env when the DB row is absent) or strike them from `.env.example`, `CLAUDE.md`, and design spec §5.4. Pick one — today docs say one thing and the code does another (finding #1, file `backend/src/domains/llm/services/llm-queue.ts:103-123`).
2. **[SHOULD FIX]** Either remove `ollamaBreakers` / `openaiBreakers` from the `/api/health` response or replace with an aggregate over `providerBreakers`. Today `/api/health.circuitBreakers.ollama.chat.state` is permanently `CLOSED` regardless of actual LLM state (finding #2, file `backend/src/routes/foundation/health.ts:149-150` + `backend/src/core/services/circuit-breaker.ts:159-194`).
3. **[SHOULD FIX]** Replace `ollamaBreakers.embed.getStatus()` / `openaiBreakers.embed.getStatus()` in `getEmbedBreakerNextRetryTime()` with a lookup keyed by the resolved embedding provider's UUID through `getProviderBreaker(providerId).getStatus()` (finding #3, file `backend/src/domains/llm/services/embedding-service.ts:126-135`).
4. **[SHOULD FIX]** Add a typed handler for `SsrfError` in the app error pipeline — or try/catch around `assertNonSsrfUrl` in the CREATE and PATCH routes — so admins see HTTP 400 with a useful reason string, not HTTP 500 (finding #4, file `backend/src/routes/llm/llm-providers.ts:47, 69`).
5. **[SHOULD FIX]** Replace the message-regex error mapping in `DELETE /admin/llm-providers/:id` with error-class or PG-error-code matching so a concurrent `PUT /llm-usecases` → FK RESTRICT race returns 409, not 500 (finding #5, file `backend/src/routes/llm/llm-providers.ts:88-96`).
6. **[SHOULD FIX]** Before merging #259 into `main`, ship #257. If #257 is too far out, add a UI warning on the embedding-dimension confirmation dialog stating "re-embed worker is not yet deployed — pages will remain unembedded until a manual re-embed is triggered". Otherwise an operator who confirms the destructive dialog will destroy retrieval for an unknown amount of time (finding #6).
7. **[CONSIDER]** Pass the real provider id to `checkLlm` instead of the `'health-probe'` sentinel, so a transient upstream hiccup doesn't trip a dedicated breaker that then stays open for 30s (finding #7).
8. **[CONSIDER]** Finish the CLAUDE.md + AGENTS.md updates promised by spec §7.4 — add deprecated markers to `LLM_PROVIDER`, `OPENAI_BASE_URL`, `OPENAI_API_KEY` in the Environment section, and update AGENTS.md to match (finding #11).
9. **[CONSIDER]** Relax `LlmProviderInputSchema.apiKey` to `.nullable().optional()` so `{apiKey: null}` is accepted as "no key" (finding #10).
10. **[CONSIDER]** Smooth the error UX between `/:id/test` and `/:id/models` — both talk to the same upstream; one returns `{connected:false, error}` with 200, the other returns 500 (finding #9).
11. **[CONSIDER]** Migrate the Add-Provider modal to `@radix-ui/react-dialog` for focus-trap and aria-labelledby (finding #12).

### Verdict: COMMENT

**Reason**: Story is PARTIALLY ADDRESSED — AC-1/2/3/4/6 are fully met with live evidence, but AC-5 is half-met (re-embed schema change works, actual page re-embedding is stubbed; author openly tracks this as #257). No CRITICAL code-review findings. 6 WARNING-level findings are worth addressing before merge: a documented env-var knob that isn't read (#1), two separate observability regressions around legacy circuit-breaker exports (#2, #3), an SSRF-error 500-instead-of-400 bug (#4), a race-prone FK-violation handler (#5), and the re-embed stub's user-visible consequence (#6). Every static check passes (typecheck, lint, 1893 backend tests, 1805 frontend tests, production build, npm audit). The refactor itself is clean — legacy services removed, domain boundaries respected, audit trail complete, ADR + architecture diagrams updated per CLAUDE.md rule 6, migration 054 is idempotent where it matters, resolver CTE is correct under concurrent writes.

---
Reviewed by PR Reviewer Agent | 2026-04-21

laboef1900:
Addressed all 6 WARNING findings in 9a56074:

1. **SsrfError → 400** — `POST`/`PATCH /admin/llm-providers` now wrap `assertNonSsrfUrl` in try/catch, catch `SsrfError`, and return `400 { error }` with no stack trace. Mirrors tests cover both verbs.
2. **DELETE race → 409** — `DELETE /admin/llm-providers/:id` now classifies PG `23503` foreign-key violations (raised when a concurrent usecase-assignment INSERT wins the row-lock race) as 409, alongside the existing message-regex path. Added a `Promise.allSettled` race test that asserts DELETE always returns 200 or 409, never 500.
3. **/api/health live breakers** — Replaced the dead `circuitBreakers.ollama` / `circuitBreakers.openai` read with a new `listProviderBreakers()` helper on `circuit-breaker.ts` that snapshots the live per-provider `Map`. Payload now exposes `circuitBreakers.providers: [{ providerId, name, state, failures }]`, enriched via `listProviders()`. Health tests cover both the tripped and orphan (breaker-without-row) cases.
4. **getEmbedBreakerNextRetryTime** — No longer reads `ollamaBreakers.embed` / `openaiBreakers.embed` (never populated by the multi-provider OpenAI-compatible client). Resolves the `embedding` use-case via `resolveUsecase()` and reads the per-provider breaker's `nextRetryTime`; returns `null` (immediate retry) when no provider is configured. Became async; callsite updated. Unit tests added for closed/tripped/reset/unresolvable paths.
5. **LLM_CONCURRENCY / LLM_MAX_QUEUE_DEPTH env vars** — `llm-queue.ts` now reads the env vars at module load as the fallback when `admin_settings` rows are absent. Precedence: `admin_settings` > env > hardcoded (4 / 50). JSDoc block at top of file documents the precedence. Env-var fallback tests exercise valid values, invalid values (NaN, 0), and a functional `LLM_CONCURRENCY=1` concurrency assertion.
6. **Re-embed dialog warning** — Heavy-path confirmation now renders a red `role="alert"` banner stating "re-embed worker not yet implemented" and linking to issue #257. Existing snapshot tests still pass; one new test asserts the alert + link.

### Verification
- `npm run typecheck`: green (contracts + backend + frontend)
- `npm run lint`: green
- `npm test -w backend`: **1907 passed** (131 files)
- `npm test -w frontend`: **1806 passed** (149 files)

No broader refactoring required; legacy `ollamaBreakers` / `openaiBreakers` globals retained for backward compat with other callsites (`llm-models.ts`, `core/index.ts` exports, several route tests).

laboef1900 (commented):
## Automated PR Review (re-review of `9a56074`)

**PR**: #259 -- feat: multi-LLM-provider configuration
**Branch**: feature/multi-llm-providers -> dev
**Diff size**: +9140 / -6145
**Reviewed at**: 2026-04-21 07:40 UTC
**Previous review**: COMMENT at `4fdbcb0` with 6 WARNING findings

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | clean after tsbuildinfo rebuild |
| Lint | PASS | only a pre-existing `boundaries/dependencies` v5→v6 migration notice |
| Backend Tests | PASS | 1907 passed / 0 failed (against test-postgres on :5433) |
| Frontend Tests | PASS | 1806 passed / 0 failed |
| Production Build | PASS | Vite bundle OK (chunk-size warnings pre-existing) |
| npm audit | PASS | 0 high / 0 critical |

### Docker App Test

| Check | Status |
|-------|--------|
| All | SKIPPED |

Docker phase skipped: port 5433 is occupied by the user's `compendiq-test-postgres-test-1` container and the running `compendiq-backend-1` is on the `dev` image, not the PR branch. A parallel stack can't claim these ports without disrupting active dev work. **Mitigating factor**: the Fastify-inject backend route tests in `backend/src/routes/llm/llm-providers.test.ts` exercise the exact endpoints with the real SSRF guard (real URL parsing / loopback detection; DNS mocked only so the mutation tests don't hang) and the real PostgreSQL — the Fix 1 / Fix 2 smoke checks requested for this pass are covered there.

### E2E Browser Tests

| Project | Status |
|---------|--------|
| All | SKIPPED |

E2E skipped: no Docker stack to target, and the Playwright spec added in this PR (`e2e/llm-providers.spec.ts`) needs a live backend running PR code.

### Verification of the 6 fixes claimed for `9a56074`

| # | Fix | Location | Verdict |
|---|-----|----------|---------|
| 1 | SsrfError → 400 on POST/PATCH | `backend/src/routes/llm/llm-providers.ts:47-54, 77-85` | **PASS**. Try/catch wraps `assertNonSsrfUrl`; only `SsrfError` is mapped to 400, everything else is re-thrown. Route test `llm-providers.test.ts:99-131` asserts 400 + `/ssrf/i` message + no stack-trace leak. |
| 2 | DELETE 23503 FK race → 409 | `backend/src/routes/llm/llm-providers.ts:104-117` | **PASS**. Classifies both the service-layer `default\|referenced` message and PG code `23503` as 409. Race test `llm-providers.test.ts:140-172` concurrently runs DELETE + usecase INSERT and asserts the DELETE response is 200 OR 409, never 500. |
| 3 | `/api/health` live per-provider breakers + names | `backend/src/core/services/circuit-breaker.ts:241-253`, `backend/src/routes/foundation/health.ts:148-172` | **PASS** (with a caveat — see finding W-2 below). `listProviderBreakers()` iterates only providers that have been touched. Health route enriches each with a human name via `listProviders()` and gracefully continues if that lookup throws. New shape is `circuitBreakers: { providers: [{providerId, name, state, failures}] }`. |
| 4 | `getEmbedBreakerNextRetryTime` async + per-provider | `backend/src/domains/llm/services/embedding-service.ts:131-141` | **PASS**. Resolves embedding use case via `resolveUsecase('embedding')`, reads breaker for the resolved provider id, returns null on resolver error. Sole caller at `:537` properly awaits. Test suite `embedding-service.test.ts:1391-1438` covers closed / tripped / reset / unresolved paths. |
| 5 | `LLM_CONCURRENCY` / `LLM_MAX_QUEUE_DEPTH` env fallback | `backend/src/domains/llm/services/llm-queue.ts:36-45` | **PASS**. `envInt()` guards against NaN / ≤0 / empty, uses `?? HARDCODED_*` fallback. `initLlmQueue()` at `:128-148` still loads `admin_settings` rows and overrides the env defaults at boot, matching the precedence documented in the file header. |
| 6 | Re-embed banner heavy path: `role="alert"` + #257 link | `frontend/src/features/settings/panels/EmbeddingReembedBanner.tsx:72-95` | **PARTIAL** — `role="alert"` and wording land correctly, but the `href` points to the wrong repo. See finding W-1 below. |

### New findings in `9a56074` (outside the original 6)

| W-1 | WARNING | `frontend/src/features/settings/panels/EmbeddingReembedBanner.tsx:86` | The #257 link introduced by Fix 6 points to `https://github.com/VAMFI/compendiq-ce/issues/257`, which returns **HTTP 404**. The real repo is `Compendiq/compendiq-ce` (verified via `gh repo view` and by fetching both URLs). Users clicking the banner's "issue #257" link during the heavy-confirm flow will hit a dead page instead of the tracking issue — which directly defeats the intent of Fix 6. The banner test at `EmbeddingReembedBanner.test.tsx:121` only asserts `href.includes('257')`, so it passes even with the wrong owner. Fix: change the href to `https://github.com/Compendiq/compendiq-ce/issues/257` and tighten the test to match the full URL. |
| W-2 | WARNING | `docs/API.md:66-69`, `docs/ADMIN-GUIDE.md:386-389` | Both docs still show the old `circuitBreakers: { ollama: ..., openai: ... }` shape in the `/api/health` example response. Fix 3 removed those keys entirely and replaced them with `circuitBreakers: { providers: [...] }` (no backward-compat shim — verified by `grep "ollama\|openai" health.ts` returning only the import line). This is API docs drift against the response the server actually returns. CLAUDE.md §Code Quality: "Every PR must include doc updates: `docs/`". |

### Acceptance Criteria (Issue #256)

Re-verified against `9a56074`; no regression vs. the `4fdbcb0` pass.

| AC-1 | Admin can create ≥2 providers and assign to any of 5 use cases | PASS | `llm-providers.ts` + `llm-usecases.ts` + `ProviderListSection.tsx` + `UsecaseAssignmentsSection.tsx`; covered by `llm-providers.test.ts` + e2e spec |
| AC-2 | Changes take effect without restart | PASS | `llm-provider-resolver.ts` — no cache on the assignment row, versioned cache on the provider config |
| AC-3 | DELETE of referenced provider → 409 | PASS | Fix 2 extends this to the concurrent-INSERT race; test `llm-providers.test.ts:140-172` |
| AC-4 | Pre-054 install preserves active provider/model | PASS | Migration `054` seeds from `admin_settings`; migration test already present |
| AC-5 | Embedding change → confirmation + re-embed | **PARTIAL** | Banner + backend endpoint ship; worker loop is explicitly a follow-up (#257). Banner warning makes the gap visible to the operator, BUT the link is broken — see W-1 |
| AC-6 | `npm test`, `npm run lint`, `npm run typecheck` pass | PASS | 3713 tests total passing, lint clean, typecheck clean |

### Story Validation (Issue #256)

**Issue intent**: Replace the two-slot (`ollama` / `openai`) LLM model with an admin-managed list of N `openai-compatible` providers + per-use-case assignments, without breaking the existing upgrade path.

**Story completion**: **FULLY ADDRESSED** (story-level). The multi-provider data model, resolver, CRUD API, use-case assignment UI, and migration seed all ship. Follow-up #257 (BullMQ re-embed worker loop) is explicitly carved out in the PR body and tracked separately.

**User journey**: Before: admin could only wire one of two hard-coded slots; embedding rode whichever slot was active. After: admin adds N providers in Settings → LLM, assigns chat / summary / quality / auto_tag / embedding to any of them, changes take effect on the next request.

**Gaps**: None at the story level. The banner copy correctly tells the operator the re-embed worker hasn't shipped yet, but W-1 means the "See issue #257 for progress" link 404s — a small but user-visible papercut inside an otherwise-complete gap-disclosure.

**Evidence**: `docs/ARCHITECTURE-DECISIONS.md` ADR-021 + new migrations `054*.sql` + `backend/src/domains/llm/services/llm-provider-{service,resolver}.ts` + `backend/src/routes/llm/llm-{providers,usecases}.ts` + `frontend/src/features/settings/panels/{ProviderListSection,ProviderEditModal,UsecaseAssignmentsSection,EmbeddingReembedBanner}.tsx`.

### Suggested Actions

1. **[SHOULD FIX]** `frontend/src/features/settings/panels/EmbeddingReembedBanner.tsx:86` — change `https://github.com/VAMFI/compendiq-ce/issues/257` to `https://github.com/Compendiq/compendiq-ce/issues/257`. Tighten `EmbeddingReembedBanner.test.tsx:121` to assert the full URL (e.g. `toBe('https://github.com/Compendiq/compendiq-ce/issues/257')`), so the wrong-owner typo can't recur.
2. **[SHOULD FIX]** `docs/API.md:66-69` and `docs/ADMIN-GUIDE.md:386-389` — replace the example `/api/health` `circuitBreakers` block with the new `providers: [{providerId, name, state, failures}]` shape. Keep them in this PR per the "every PR includes doc updates" rule.
3. **[OPTIONAL]** Consider a small integration assertion that `GET /api/health` response schema does not contain the legacy `ollama` or `openai` keys — would have caught W-2 automatically.

### Verdict: COMMENT

**Reason**: All 6 originally-flagged WARNINGs are correctly addressed in `9a56074` with matching tests. However, Fix 6 introduces a dead 404 link in the very banner the fix was supposed to improve (W-1), and the PR still leaves the `/api/health` shape documented in two doc files (W-2) against the new code. Neither blocks merge, but both are trivial to correct and directly in the fix-6 / fix-3 scope. Labels: remove `PR:Warning`, keep unresolved — neither `PR:Approved` nor `PR:Critical`. Once W-1 + W-2 are addressed, this is a clean APPROVE.

---
Reviewed by PR Reviewer Agent | 2026-04-21

laboef1900:
**Verdict correction: APPROVE**

Per re-review instructions ("new warning-level findings shouldn't block approval unless severe"), both new findings above (broken `VAMFI` link in the re-embed banner, stale `/api/health` shape in API docs) are polishable but not severe — the original 6 WARNING fixes are all correctly in place with matching tests. Labels updated: `PR:Warning` removed, `PR:Approved` added.

Recommend folding W-1 + W-2 into a small follow-up commit on this branch before merging, but they do not block the merge.

PR #317, opened 2026-04-23
feat(#302 Gap 3): flush pending draw.io diagrams to attachments before save

laboef1900:
Closes **#302 Gap 3**. Builds on the Gap 2 route landed in PR #316. Part of epic #295 (Phase 1.2 CE scope).

## Problem

`DrawioDiagramNodeView.handleSave` writes the edited PNG + XML into the node's transient `pngDataUri` / `xml` attributes. The page-level save in `PageViewPage.handleSave` never flushes those to the attachment store — it just calls `updateMutation.mutateAsync({ bodyHtml: editHtml })`. `DrawioDiagram.renderHTML` then serialises the PNG as ``.

Visually this works in the editor. But:
- `body_html` bloats by ~14 MB per edit (the PNG ships base64-encoded).
- The node's `src` attribute stays null, so the next edit-and-save cycle bloats the row **again** — there's no convergence.
- On reload the DB row carries a huge inline data URI instead of a short attachment URL.

## Fix

New `frontend/src/shared/components/article/drawio-save-drain.ts`. Called from `PageViewPage.handleSave` **before** the update mutation:

1. Walks `editor.state.doc.descendants` looking for `drawioDiagram` nodes with `pngDataUri` set and no `src`.
2. For each, `PUT /api/attachments/:pageId/:filename` (PNG body + XML sibling per Gap 2 route).
3. Rewrites the node via ProseMirror `setNodeMarkup(pos, null, { …attrs, src: resolvedUrl, pngDataUri: null })` so `editor.getHTML()` emits the short server URL.
4. `handleSave` then grabs the post-drain HTML via `editor.getHTML()` instead of the pre-drain `editHtml` state.

Standalone (non-Confluence) pages currently skip with a `toast.warning` — they need a new attachment backend tracked in #302 Gap 4.

## Tests

New `drawio-save-drain.test.ts` with 6 cases:
- no-op when editor is null
- no-op when no diagrams have a pending `pngDataUri`
- happy path: two pending diagrams upload, `src` rewritten, `pngDataUri` cleared, XML passed when present
- standalone-page path skips with the Gap 4 error message
- per-diagram failure isolated (one boom + one success → 1 uploaded / 1 failed)
- missing `attachmentPageId` fails cleanly

Stubs the TipTap Editor shape (state.doc.descendants, nodeAt, state.tr.setNodeMarkup, view.dispatch) without pulling in the full ProseMirror runtime. `fetch` is mocked at the global level.

[...]

PR #316, opened 2026-04-23
feat(#302 Gap 2): round-trip draw.io XML with Confluence's native viewer

laboef1900:
Closes #302 Gap 2 (the highest-value gap). Part of epic #295 (Phase 1.2 CE scope, §3.17 Editor Improvements).

## Scope-reality note up front

The #302 issue body was written against an older state of the codebase. **Gap 1 (self-hosted draw.io container)** was already solved in v0.3 — `frontend/Dockerfile` bundles `draw.war` v29.6.4 into the nginx image under `/usr/share/nginx/html/drawio/` and `frontend/nginx.conf` already serves it at `/drawio/`. No sidecar needed, no reverse-proxy CORS concerns. The issue body should be updated.

**This PR ships Gap 2 only** — the highest-value remaining gap. Gaps 3 and 4 remain open; see "Follow-ups" below.

## Summary (Gap 2)

`frontend/src/features/pages/PageViewPage.tsx:272` had `handleDrawioSave(dataUri, _xml)` — the `_xml` prefix flagged the arg as unused, and it was. The PNG went to Confluence, the XML was dropped. Result: a diagram edited in Compendiq and synced back to Confluence was **no longer editable in Confluence's native draw.io viewer** because the viewer reads the `.drawio` XML sibling attachment, not the PNG.

Changes:

1. **Backend** (`routes/confluence/attachments.ts`):
   - `UpdateAttachmentBodySchema` gains an optional `xml` field, max 25 MB.
   - `PUT /api/attachments/:pageId/:filename` uploads the PNG as before, then uploads the XML (when provided) as a sibling `.drawio` attachment with content-type `application/xml`.
   - Filename derivation: `diagram.png → diagram.drawio`; non-PNG filenames append `.drawio` (`custom.svg → custom.svg.drawio`) to match Confluence's own plugin convention.
   - XML upload failure is **non-fatal** — the PNG has already landed, so the response still reports success (XML failure is logged as a warning). This lets the UI stay working when Confluence rejects XML content for some reason.
   - Response body gains optional `xmlFilename` and `xmlSize` fields.

2. **Backend** (`domains/confluence/services/attachment-handler.ts`):
   - `syncDrawioAttachments` now downloads the `.drawio` XML sibling alongside the PNG when Confluence has one. Missing `.drawio` is not an error (older pages without XML siblings continue to work — view-mode needs only the PNG).
   - Restructured the existing PNG-cache check so the `.drawio` sibling check runs independently of the PNG cache-skip branch.

[...]

Issue #302, opened 2026-04-22
[phase-1.2] Draw.io inline editing — self-hosted server + XML round-trip + local-page storage


laboef1900:
**Parent epic:** #295 (Phase 1.2 CE scope, §3.17 Editor Improvements).

## Important framing — what already ships

The codebase research (gemini agent against the live frontend + backend) reveals the Draw.io feature is **substantially further along than the §3.17 description implies**. Treat this as a **gap-closure issue**, not a greenfield design.

### Already shipping in v0.3

- `frontend/src/shared/components/article/DrawioEditor.tsx` — full-screen iframe overlay with the complete postMessage protocol (init → load → save → export xmlpng → exit)
- `frontend/src/shared/components/article/DrawioDiagramNodeView.tsx` — TipTap v3 NodeView wrapping `DrawioEditor`, double-click to edit, drag handle, hover overlay
- `DrawioDiagram` TipTap node in `article-extensions.ts` with `xml` / `pngDataUri` / `src` / `diagramName` attrs, `insertDrawioDiagram` toolbar command (`Editor.tsx:423`)
- `PUT /api/attachments/:pageId/:filename` — uploads edited PNG back to Confluence via the user's PAT, writes local cache, invalidates Redis
- `GET /settings/drawio-url` + admin `PUT` — stores configurable embed URL in `admin_settings`; defaults to `https://embed.diagrams.net` when unset
- `ArticleViewer.tsx` — injects inline edit buttons on read-only view, calls `onEditDiagram` → `handleEditDiagram` in `PageViewPage`
- Embed URL already uses `offline=1&stealth=1` to disable cloud integrations

### Gap 1: No self-hosted draw.io container in `docker-compose.yml`

The fallback to `https://embed.diagrams.net` phones home — **a hard no for enterprise/on-prem customers**. The `DrawioEditor.tsx:224` error message even refers to "nginx container" that does not exist yet in the compose file.

**Fix:**
- Add `drawio` service to `docker/docker-compose.yml` using `jgraph/drawio:26.x` (Apache 2.0; pin a specific full-version tag like `26.0.4` — `latest` has broken embeds across minor versions)
- Internal network only, no external port exposure
- Reverse-proxy `/drawio/` through the existing frontend nginx so the iframe `effectiveOrigin` check matches the frontend origin (avoids same-origin / CORS issues with the postMessage protocol)
- Update `docs/architecture/05-deployment.md` Mermaid to show the new sidecar
[...]

Issue #295, opened 2026-04-22
[phase-1.2 epic] v0.4 Polish & Ecosystem — Community Edition scope


laboef1900:
## Phase 1.2 / v0.4 — CE epic

**Target:** 2026-07-01 (~10 weeks after v0.3 shipped 2026-04-22)
**Source of truth:** `Documentation-Research/Compendiq/Release Roadmap.md` §3.17 + §3.20 (post-2026-04-22 scope trim)
**Companion EE epic:** Compendiq/compendiq-ee#110

This issue tracks the **Community-Edition** half of v0.4. Items here ship in the open repo (AGPL-3.0). Pure-Enterprise items (webhooks, per-space ACLs, multi-instance, etc.) are tracked in the EE epic.

Sub-issues filed 2026-04-22 with research-backed scope. Prerequisite issues added 2026-04-22 after the gh-issue-reviewer pass discovered four CE-side gaps that gate EE work in the companion epic.

---

### Carry-overs from v0.3.0 closeout

- [ ] #30 — UI premium light/dark themes *(already an open CE issue; no new issue needed)*

### §3.17 Editor Improvements

- [ ] #302 — **Draw.io inline editing**: self-hosted server + XML round-trip + local-page storage
- [ ] #301 — **Real-time co-presence indicators**: ghost avatar stack via Redis sorted-set + SSE
- [ ] #300 — **Improved paste-from-Confluence handling**: JIRA + include + @mention + TOC macros

### §3.19 Cross-cutting (CE-side implementation)

- [ ] #303 — **PDF dashboard upgrade + Excel removal** *(rescoped 2026-04-22 — current PDF is minimal and Excel was dropped in trim)*

### §3.20 Community Growth

- [ ] #296 — **Public stewardship commitment** (`docs/STEWARDSHIP.md`)
- [ ] #297 — **Public roadmap page** (GitHub Projects v2 board)
- [ ] #298 — **Case studies** from v0.2/v0.3 enterprise customers *(Phase A templates must ship; Phase B published case studies may slip to v0.5)*
- [ ] #299 — **Integration guides** for the three Confluence DC deployment shapes (reverse proxy / self-signed TLS / air-gapped)

### CE prerequisites for EE work (added 2026-04-22 from gh-issue-reviewer findings)

These are CE-side gaps the review pass discovered. Each gates a specific EE issue, but each is also valuable independently — most either close latent bugs or polish under-served admin UX.

[...]

PR #318, opened 2026-04-23
feat(#302 Gap 4): local-page attachment storage

laboef1900:
Closes **#302 Gap 4**. Builds on #316 (Gap 2, merged XML round-trip) and #317 (Gap 3, TipTap drain helper — currently skips standalone pages). Part of epic #295 (Phase 1.2 CE scope).

## Summary

Standalone (non-Confluence) pages had nowhere to persist diagram blobs — `/api/attachments/…` requires a Confluence page and filesystem cache keyed by `confluence_id`. This PR adds a parallel attachment store rooted at the local page's numeric PK.

### Migration 063

New `local_attachments` table:
- `page_id INT` FK to `pages(id)` with `ON DELETE CASCADE` (so trashing a page takes its attachments with it)
- `filename TEXT`, `content_type TEXT`, `size_bytes BIGINT`, `sha256 TEXT`
- `created_by UUID` FK to `users(id)` with `ON DELETE SET NULL`
- `UNIQUE(page_id, filename)` for upsert semantics + index on `page_id` for fast list queries

### Service layer

`core/services/local-attachment-service.ts`:
- Typed `LocalAttachmentError` with codes mapped to HTTP at the route layer (`NOT_FOUND` → 404, `PAGE_NOT_FOUND` → 404, `FORBIDDEN` → 403, `TOO_LARGE` → 413, `INVALID_FILENAME` → 400)
- `putLocalAttachment` / `getLocalAttachment` / `listLocalAttachments`
- **Access control** via `assertLocalPageAccess`:
  - Rejects Confluence-backed pages (forces the Confluence route)
  - Rejects trashed pages
  - `visibility = 'private'`: owner only
  - `visibility = 'shared'`: any authenticated user
- Path traversal guard: `path.basename()` on filenames, `resolve() + prefix check` on directory paths, explicit hidden-filename rejection (`.secret`)
- 25 MB size cap

### Routes (`/api/local-attachments/:pageId/…`)

- `GET list` — `{attachments: [{filename, size, contentType, sha256, updatedAt, url}]}`
- `GET :filename` — raw bytes with `Cache-Control: private, max-age=3600`; SVGs additionally carry `Content-Security-Policy: sandbox` and `Content-Disposition: attachment` to match the Confluence-side hardening
- `PUT :filename` — accepts `{dataUri, xml?}` (same shape as the Confluence route). When `xml` is supplied, persists a matching `.drawio` sibling (parity with #302 Gap 2)

### Filesystem layout

`ATTACHMENTS_DIR/local/<page_id>/<filename>` — separate subtree from Confluence's `ATTACHMENTS_DIR/<confluence_id>/…` so the two namespaces never collide.

### Follow-up

The **Gap 3 drain helper** (frontend `drawio-save-drain.ts` in PR #317) currently skips standalone pages with a friendly toast. After this PR and #317 both merge, a tiny follow-up will swap the URL prefix in `uploadOne` so the drain routes to the local backend when `pageSource === 'standalone'`. Deliberately not included here — kept the PR focused on the backend half.

## Tests

`core/services/local-attachment-service.test.ts` — 13 integration tests against real Postgres:
- column + constraint + index shape
- upsert semantics (second PUT bumps `sha256` + `updated_at`, same row id)
- get returns bytes + metadata
- list sorts filenames
- forbid non-owner on private page; allow non-owner on shared page
- reject Confluence-backed page with FORBIDDEN
- reject trashed page with PAGE_NOT_FOUND
- reject oversized payload with TOO_LARGE
- reject hidden / traversal filenames with INVALID_FILENAME
- `NOT_FOUND` when DB row present but file missing (storage drift)
- `ON DELETE CASCADE` wipes attachments when the page is deleted

Full backend: 141 files / 2053 passing / 1 skipped. Lint + typecheck green.

## Test plan

- [x] Migration applies cleanly
- [x] Every service branch covered by a test
- [x] Security guards: path traversal, size cap, hidden filenames, Confluence-page rejection
- [x] SVG CSP + disposition parity with the Confluence attachment route
- [ ] CI green
- [ ] Manual smoke: create a standalone page, paste an image → confirm it persists across reload (needs the #317 drain-helper wiring follow-up to actually be exercised end-to-end)


laboef1900 (commented):
## Automated PR Review

**PR**: #318 -- feat(#302 Gap 4): local-page attachment storage
**Branch**: feature/302-gap4-local-attachments -> dev
**Diff size**: 836 lines (+765/-0)
**Reviewed at**: 2026-04-23 08:16 UTC

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | |
| Lint | PASS | backend + frontend clean (one pre-existing `boundaries` v5→v6 deprecation warning unrelated to this PR) |
| Backend Tests | PASS | 2053 passed, 1 skipped (141 files) |
| Frontend Tests | PASS | 1822 passed (150 files) |
| Production Build | PASS | Vite build clean (only the usual >600kB chunk-size warnings, pre-existing) |
| npm audit | PASS | 0 high / 0 critical (4 moderate in `uuid` via bullmq/exceljs/mermaid, pre-existing, below threshold) |

### Docker App Test

| Check | Status |
|-------|--------|
| Login | SKIPPED |
| Backend /health | SKIPPED |
| API /dashboard | SKIPPED |
| API /containers | SKIPPED |
| Frontend loads | SKIPPED |
| Migrations | SKIPPED |
| No error logs | SKIPPED |

Docker tests skipped: ports 5433 (test Postgres) and 6379 (Redis) were in use by the concurrent dev stack. Migration 063 did apply cleanly against the running test Postgres during the backend integration suite (all 13 new service-layer tests passed, which includes the `information_schema.columns` shape assertion for `local_attachments`).

### E2E Browser Tests

| Project | Status |
|---------|--------|
| auth | SKIPPED |
| setup | SKIPPED |
| chromium | SKIPPED |

E2E tests skipped: Docker app stack not available (see above). No existing E2E specs target local attachments; the Gap 3 drain wiring that would exercise this route end-to-end is deliberately out of scope (see PR body).

### Code Review Findings

| 1 | WARNING | backend/src/routes/knowledge/local-attachments.ts:135 | PUT route does not set a per-route `bodyLimit`. The advertised 25 MB size cap (service layer) and the route-layer 413 branch at line 144 are both unreachable: Fastify's default JSON body limit is 1 MB, so any payload between ~1.3 MB and 25 MB is rejected with Fastify's generic `FST_ERR_CTP_BODY_TOO_LARGE` 413 before the handler runs. Set `{ bodyLimit: 35_000_000 }` (base64 inflates by ~33%, plus small JSON overhead) on the route options, mirroring the pattern in `pages-crud.ts:1665` (`{ bodyLimit: 15_000_000 }`). Note: the Confluence-side `PUT /api/attachments/:pageId/:filename` has the same latent bug with its 10 MB cap — worth fixing both together, but calling it out here because Gap 4's whole point is enabling standalone-page diagrams, and 1 MB rules out most non-trivial PNG+XML pairs. |
| 2 | WARNING | backend/src/routes/knowledge/local-attachments.ts:1-201 | No HTTP-level integration tests. The service layer has excellent coverage (13 tests), but the route-specific logic is entirely untested: dataUri parsing (`decodeDataUri` regex + MIME allowlist at line 51), the 413-before-413 defensive check (line 144, unreachable as noted above), the SVG CSP + content-disposition branch (lines 111-117), the XML sibling filename transform (lines 166-168), the error→status mapping (`mapErrorToStatus`), and the "XML sibling write failed but PNG stored" silent-success path (lines 177-183). `local-spaces.test.ts` exists as a template for how route-layer tests are done in this repo. |
| 3 | WARNING | backend/src/routes/knowledge/local-attachments.ts:165-183 | XML sibling write silently swallows errors. When `body.xml` is supplied and the XML write fails (DB constraint, disk full, invalid xml filename), the handler logs a `warn` and returns `{success: true, xmlFilename: undefined, xmlSize: undefined}`. The caller (drain helper) has no way to distinguish "no XML supplied" from "XML write failed", and the on-disk state diverges from Confluence parity (PNG without sibling .drawio). Consider either (a) returning an explicit `xmlWriteFailed: true` flag, or (b) failing the whole PUT with 500 so the client retries. Option (a) is probably right here given the follow-up drain helper can re-send. |
| 4 | INFO | backend/src/core/services/local-attachment-service.ts:68 | Path traversal guard uses `resolved.startsWith(baseResolved + path.sep) \|\| resolved === baseResolved`. This is correct on POSIX but `path.sep` is `\` on Windows — if the project ever runs the backend on Windows, `baseResolved + '/'` would fail. Non-issue for Linux-only Docker deployments; noting for completeness. Additionally, `path.basename('..\\secret')` returns `..\\secret` verbatim on Linux (backslash isn't a separator), which would then trip the `.startsWith('.')` check and get rejected — safe but by the secondary guard, not the primary one. |
| 5 | INFO | backend/src/core/services/local-attachment-service.ts:31 | `attachmentsBase()` resolves `process.env.ATTACHMENTS_DIR` at every call. The docstring justifies this for test override flexibility, and the perf cost is negligible, but the Confluence route caches it at module load (`const ATTACHMENTS_BASE = process.env.ATTACHMENTS_DIR ?? …` in `confluence/attachments.ts:18`). Minor inconsistency — the new approach is actually better (supports hot-reload in tests), just flagging the divergence in case consistency is preferred. |
| 6 | INFO | backend/src/core/db/migrations/063_local_attachments.sql | Migration number jumps from 057 to 063 with no explanation. PRs #310 (060), #311 (061), #312 (062) are open and reserve those slots; 058 and 059 are currently free. Not a bug (migrations sort lexically), but a comment at the top of the SQL noting the reservation strategy would help whoever lands 058/059 in the future. |
| 7 | INFO | backend/src/core/services/local-attachment-service.ts:143 | Access control mirrors `pages-crud.ts`'s "owner OR shared" rule but the comment on line 141 mentions admins "pass through (the route layer has separate requireAdmin-style plumbing if needed)". There is no admin bypass in the current service, and the route does not add one — so an admin currently cannot read/write attachments on another user's private page. That may be intentional (least privilege), but the comment implies otherwise. Either add the admin bypass if that's the design intent, or tighten the comment. |

### Acceptance Criteria (Issue #302 Gap 4)

Gap 4 is one of four gaps in issue #302. The gap-specific requirement (from issue body):

| AC-1 | Store diagram blobs in Compendiq filesystem under `ATTACHMENTS_DIR` using synthetic page-id scheme keyed by Compendiq page id (option A) | PASS | `local-attachment-service.ts:23,63-72` — files at `ATTACHMENTS_DIR/local/<page_id>/<filename>`, separate subtree from Confluence cache |
| AC-2 | New `local_attachments` table: page_id FK, filename, content_type, size, sha256, created_at, created_by | PASS | `migrations/063_local_attachments.sql:16-28` — all columns + `updated_at` bonus, FK to `pages(id) ON DELETE CASCADE`, FK to `users(id) ON DELETE SET NULL` |
| AC-3 | Filesystem path resolution (new routes for local attachments) | PASS | `routes/knowledge/local-attachments.ts` — GET list / GET one / PUT, wired in `app.ts:297` |
| AC-4 | Parent-issue AC "Local-page diagram editing works end-to-end (insert via toolbar → edit → save page → reload → diagram still there)" | NOT IN SCOPE | PR body explicitly defers this to a follow-up that swaps the URL prefix in the Gap 3 drain helper's `uploadOne`. This is the backend half only; end-to-end verification requires #317 + follow-up. |

### Story Validation (Issue #302 Gap 4)

**Issue intent**: Give standalone (non-Confluence) pages a place to persist draw.io diagrams and pasted images. The existing `/api/attachments` route is keyed on Confluence IDs and per-user PATs, so local pages had no attachment store at all — `insertDrawioDiagram` created diagrams that were orphaned on save. Option A in the issue: a parallel filesystem tree + DB table keyed on the Compendiq page PK.

**Story completion**: FULLY ADDRESSED (for the backend scope the PR claims)

**User journey**:
- Before: Standalone page in editor → insert draw.io → edit → save page → PNG blob in `pngDataUri` node attr → renderHTML drops it → reload → diagram gone.
- After (this PR only): Backend can receive PUT requests at `/api/local-attachments/:pageId/:filename` with `{dataUri, xml?}`, persists PNG + optional .drawio XML sibling under `ATTACHMENTS_DIR/local/<pageId>/`, enforces owner/shared access, rejects Confluence-backed pages, returns URLs for GET. The drain-helper wiring in frontend (`drawio-save-drain.ts`) remains in PR #317 and the follow-up to route `standalone` pages here is explicitly punted — which is the right scope call for this PR.
- The full end-to-end user journey (insert → edit → save → reload → present) still requires PR #317 + the wiring follow-up. This PR is the first of two/three steps.

**Gaps**: Scope deliberately limited to backend half. The follow-up to change `uploadOne` URL prefix in the Gap 3 drain helper is called out in the PR body ("Deliberately not included here — kept the PR focused on the backend half") and is the right call for reviewability.

**Evidence**: Migration + service + route + 13 integration tests all land in one coherent chunk; every branch of the service (upsert, owner/shared/forbidden, confluence-rejection, trashed, too-large, invalid-filename, file-missing, cascade-delete) has a test; SVG CSP + disposition parity with the Confluence route is preserved; DB shape matches the issue's option-A proposal plus one pragmatic addition (`updated_at` for upsert tracking).

### Suggested Actions

1. **[SHOULD FIX]** Add `{ bodyLimit: 35_000_000 }` to the PUT route registration (`local-attachments.ts:135`) so the advertised 25 MB cap is actually reachable. Without this, any PNG+XML pair over ~750 KB gets rejected with Fastify's generic 413 before the route logic runs. Same fix probably wanted in `confluence/attachments.ts:229` for the existing 10 MB cap.
2. **[SHOULD FIX]** Add a route-layer test file `local-attachments.test.ts` covering: dataUri decoding (valid + invalid MIME + non-data-URI), SVG CSP header, error→status mapping (403/404/413/400), and the XML sibling filename transform + failure path. `local-spaces.test.ts` is a good template.
3. **[CONSIDER]** Surface XML-sibling write failures to the caller instead of silently returning `success: true, xmlFilename: undefined` — either an explicit `xmlWriteFailed` flag or a hard 500 so the drain helper can retry.
4. **[CONSIDER]** Decide whether admins should have a bypass on private-page attachments and either add it (matching the comment at `local-attachment-service.ts:141`) or tighten the comment.
5. **[OPTIONAL]** Document the migration-number gap (057 → 063) in a comment at the top of `063_local_attachments.sql`, or renumber to 058 if the slots-reserved-by-other-open-PRs convention isn't intentional.

### Verdict: COMMENT

**Reason**: Implementation is clean, well-documented, and the service layer is thoroughly tested (13 integration tests covering every branch including the subtle "DB row present but file missing" drift case and `ON DELETE CASCADE`). Acceptance criteria for Gap 4 are fully met and the scope is appropriately tight (backend half only, frontend drain wiring deferred to a follow-up). However, the 25 MB body-limit advertisement is unreachable under Fastify's 1 MB default — a merge blocker for the advertised functionality, though not a correctness bug in the code as written. Combined with the missing route-layer tests, this is a COMMENT rather than APPROVE.

---
Reviewed by PR Reviewer Agent | 2026-04-23

laboef1900:
## Review fixes applied

Commit: df33499 (pushed to `feature/302-gap4-local-attachments`)

### Fixes

- **Finding #1 (bodyLimit unreachable)** — added `{ bodyLimit: 35_000_000 }` to `PUT /api/local-attachments/:pageId/:filename` and `{ bodyLimit: 15_000_000 }` to `PUT /api/attachments/:pageId/:filename`. Both in-handler size caps (25 MB local / 10 MB Confluence) are now reachable. Each got a new test that PUTs a 2 MB payload and asserts success, proving the cap has been raised above Fastify's 1 MB default.
- **Finding #2 (no HTTP route tests)** — new `backend/src/routes/knowledge/local-attachments.test.ts` with 16 tests covering dataUri parsing (valid png/svg, non-data-URI, invalid MIME shape), SVG `Content-Security-Policy: sandbox` + `Content-Disposition: attachment` hardening (and the negative PNG case), XML sibling `.drawio` filename transform + write, error→status mapping for all four `LocalAttachmentError` codes + GET 404, the >2 MB bodyLimit test, and URL-encoded filenames in the list response.
- **Finding #3 (silent XML failure)** — chose reviewer's option (a). When `body.xml` is supplied and the XML write throws, the route now returns `success: false`, `xmlWriteFailed: true`, and `xmlWriteError` (the `LocalAttachmentError.code` or the raw error message). The PNG is still persisted (which is the useful partial outcome), but the drain helper can now detect the failure and re-send the XML half. Two tests cover this (generic Error + `LocalAttachmentError`).

### Deferred

- Findings #4-7 (INFO-level: Windows `path.sep`, `attachmentsBase()` caching divergence, migration number gap, admin bypass comment) — not addressed in this PR per the "skip unless trivial" rule. Happy to pick them up in a follow-up.

### Verification

- `npm run typecheck` — pass
- `npm run lint` — pass (one pre-existing `boundaries` v5→v6 warning, unrelated)
- `npm run test -w backend` — 2070 passed / 1 skipped (142 files), up from 2053 passed at initial review — the +17 new tests are the route-layer coverage above.

Issue #352, opened 2026-04-27
Space home: "Show home content" broken; allow setting home folder; hide home folder from nav


laboef1900:
## Problem

Three related issues with the **Space home** behavior:

1. **"Show space home content" is not working.** When opening a space, the configured home content does not render.
2. **No way to set the home folder.** There is currently no UI/setting to choose which folder (or page) acts as the space's home.
3. **Home folder shown in the nav bar wastes space.** The home folder appears as its own entry in the navigation tree, which is redundant — the space already has a "Home" entry — and it eats vertical space in the sidebar.

## Expected behavior

1. **Home content renders.** Opening a space (or clicking the space's "Home" entry) shows the configured home folder/page content reliably.
2. **Settable home.** Admins (or the space owner, depending on RBAC) can pick any folder/page in the space as the home target, via the space settings UI. The choice is persisted and used by the home view.
3. **Hidden in nav.** The folder/page chosen as the home is **not** rendered as a separate item in the side navigation tree — it's only reachable via the space's "Home" entry.

## Acceptance criteria

- [ ] Root cause of the broken "show space home content" view identified and fixed
- [ ] New setting: "Home folder/page" picker in space settings (folder or page selectable from within the space)
- [ ] Setting persists in the database (add migration if needed; update `docs/architecture/06-data-model.md` per CLAUDE.md rule 6)
- [ ] Backend exposes the setting via the existing space settings contract in `@compendiq/contracts`
- [ ] Side navigation tree filters out the entry that is currently designated as the space home
- [ ] Space "Home" link in the UI uses the new setting and falls back to a sensible default when none is set (document the fallback)
- [ ] Permissions / RBAC respected: the home folder is only hidden in nav for users who can see it via the "Home" entry; if a user lacks access to the configured home, fall back to the default
- [ ] Tests: backend test for the setting persistence and resolution; frontend test for nav filtering and home rendering
- [ ] Works in both themes (Graphite Honey + Honey Linen)

## Open questions

- Should the home target be **folder-only**, **page-only**, or **either**? Recommendation: either — many users will want a single landing page, others a folder overview.
- What's the default when no home is set? (e.g. first top-level folder, or a generated overview page.)
- Who can change the setting — admin only, space owner, or any space member? Tie this to the existing RBAC model.

## References

- Frontend: `frontend/src/features/spaces/`, sidebar/nav components
- Backend: space settings route + service
- Contracts: `packages/contracts` (space settings schema)
[...]

PR #377, opened 2026-04-27
feat(#352): per-space custom home page + admin/space-owner RBAC

laboef1900:
## Summary
Three-in-one fix for the space home page:

### 1. Settable home
Was: only the Confluence-derived `space.homepage` was usable. Migration 071 adds `spaces.custom_home_page_id INTEGER REFERENCES pages(id) ON DELETE SET NULL` so admins or space owners can override the Confluence default — including pointing at a folder or shared standalone page.

### 2. Auth model
`PUT /api/spaces/:key/home` is gated on the existing `userHasPermission('manage', spaceKey)` check. That returns true for system admins (early-return in `rbac-service.ts:112`) and for anyone holding the `space_admin` system role on this space (direct or via group). RBAC seed from migration 039 covers the "space owner" case without a new concept. Page-existence and same-space-or-shared-standalone checks reject smuggling an inaccessible page as home.

### 3. Hidden in nav
Was: configured home appeared as its own tree entry, wasting a slot. `buildTree` now drops the homepage from the visible roots and promotes its children to top-level so the rest of the tree stays navigable. The home stays reachable via the dedicated space "Home" link.

`GET /api/spaces` returns the resolved home as before — `homepageId` now prefers `custom_home_page_id` when set, else falls back to the Confluence-derived id. New `customHomePageId` field surfaces the raw override for the eventual settings UI.

## Out of scope (deferred follow-up)
**The home picker UI in space settings.** Backend + data layer ship now; the picker is a focused follow-up that hits the existing PUT endpoint. Issue #352 is partially closed by this PR — the "settable home" half — and the open piece is just the UI to call PUT.

Closes part of #352 · architecture diagram updated (CLAUDE.md rule 6)

## Test plan
- [x] backend spaces suite: 8/8 (GET shape, override visibility, PUT 403/200/400/null/404)
- [x] frontend SidebarTreeView suite: 53/53 (homepage hidden + children promoted + other roots visible)
- [x] backend + frontend typecheck clean
- [ ] manual: as space-admin user, call `PUT /api/spaces/MYKEY/home` with `{ homePageId: N }` → confirm 200 + GET reflects custom id
- [ ] manual: as plain user, same call → confirm 403
- [ ] manual: with custom home configured, navigate to that space → confirm home page no longer appears as a tree entry but is reachable via the Home link

### DIFF
--- a/backend/src/core/db/migrations/071_spaces_custom_home_page.sql
+++ b/backend/src/core/db/migrations/071_spaces_custom_home_page.sql
@@ -1,0 +1,23 @@
+-- #352: per-space custom home page.
+--
+-- Today, `spaces.homepage_id` (TEXT) stores the Confluence-derived home page
+-- (whatever `space.homepage.id` was at last sync). We add a separate
+-- `custom_home_page_id` (INT FK → pages.id) that admins or space owners can
+-- set to override the Confluence default — including pointing at a folder or
+-- a standalone page that isn't part of any synced Confluence space.
+--
+-- Resolution at read time:
+--   1. If custom_home_page_id IS NOT NULL → use that page.
+--   2. Else fall back to the Confluence-derived homepage_id → pages.id lookup
+--      (the existing JOIN in routes/confluence/spaces.ts already does this).
+--
+-- ON DELETE SET NULL so deleting the configured home page doesn't break the
+-- space; resolution falls through to the Confluence default.
+
+ALTER TABLE spaces
+  ADD COLUMN IF NOT EXISTS custom_home_page_id INTEGER
+    REFERENCES pages(id) ON DELETE SET NULL;
+
+CREATE INDEX IF NOT EXISTS idx_spaces_custom_home_page_id
+  ON spaces(custom_home_page_id)
+  WHERE custom_home_page_id IS NOT NULL;
--- a/backend/src/core/services/redis-cache.test.ts
+++ b/backend/src/core/services/redis-cache.test.ts
@@ -709,3 +709,59 @@ describe('RedisCache.getOrCompute', () => {
     expect(computeFn).toHaveBeenCalledOnce();
   });
 });
+
+describe('RedisCache.invalidateAcrossUsers (#352)', () => {
+  let mockRedis: ReturnType<typeof createMockRedisClient>;
+  let cache: RedisCache;
+
+  beforeEach(() => {
+    mockRedis = createMockRedisClient();
+    cache = new RedisCache(asRedis(mockRedis));
+  });
+
+  it('SCANs the cross-user pattern and deletes every matching key', async () => {
+    // One SCAN page, two matching keys across two different userIds.
+    mockRedis.scan.mockResolvedValueOnce({
+      cursor: '0',
+      keys: ['kb:alice:spaces:list', 'kb:bob:spaces:list'],
+    });
+    mockRedis.del.mockResolvedValue(2);
+
+    await cache.invalidateAcrossUsers('spaces');
+
+    expect(mockRedis.scan).toHaveBeenCalledWith('0', {
+      MATCH: 'kb:*:spaces:*',
+      COUNT: 100,
+    });
+    expect(mockRedis.del).toHaveBeenCalledWith([
+      'kb:alice:spaces:list',
+      'kb:bob:spaces:list',
+    ]);
+  });
+
+  it('walks the cursor across multiple SCAN pages', async () => {
+    mockRedis.scan
+      .mockResolvedValueOnce({ cursor: '42', keys: ['kb:alice:spaces:list'] })
+      .mockResolvedValueOnce({ cursor: '0', keys: ['kb:bob:spaces:list'] });
+    mockRedis.del.mockResolvedValue(1);
+
+    await cache.invalidateAcrossUsers('spaces');
+
+    expect(mockRedis.scan).toHaveBeenCalledTimes(2);
+    expect(mockRedis.del).toHaveBeenCalledTimes(2);
+  });
+
+  it('does not call DEL when SCAN returns no keys', async () => {
+    mockRedis.scan.mockResolvedValueOnce({ cursor: '0', keys: [] });
+
+    await cache.invalidateAcrossUsers('spaces');
+
+    expect(mockRedis.del).not.toHaveBeenCalled();
+  });
+
+  it('soft-fails when SCAN throws (logs, does not raise)', async () => {
+    mockRedis.scan.mockRejectedValue(new Error('Redis down'));
+
+    await expect(cache.invalidateAcrossUsers('spaces')).resolves.toBeUndefined();
+  });
+});
--- a/backend/src/core/services/redis-cache.ts
+++ b/backend/src/core/services/redis-cache.ts
@@ -491,6 +491,27 @@ export class RedisCache {
   }
 
   /**
+   * Invalidate every user's cache of the given type.
+   *
+   * Used when an admin/space-owner mutation changes data that is visible to
+   * all users (e.g. setting a space's custom home page — #352). The per-user
+   * `invalidate(userId, type)` only clears the calling admin's cache and
+   * leaves every other user reading stale data for up to TTL seconds.
+   *
+   * Implemented as a single SCAN cursor walk over `kb:*:{type}:*` against the
+   * shared Redis. Works in single-pod and multi-pod deployments without
+   * needing pub/sub — Redis is the authoritative store, so deleting the keys
+   * is sufficient; subsequent reads from any pod miss and re-populate.
+   */
+  async invalidateAcrossUsers(type: CacheType): Promise<void> {
+    try {
+      await this.scanAndDelete(`kb:*:${type}:*`);
+    } catch (err) {
+      logger.error({ err, type }, 'Redis cache invalidate across users error');
+    }
+  }
+
+  /**
    * Cache-aside with stampede protection.
    *
    * On cache miss, acquires a short-lived NX lock so only one caller computes
--- a/backend/src/core/services/webhook-outbox-poller.test.ts
+++ b/backend/src/core/services/webhook-outbox-poller.test.ts
@@ -494,6 +494,17 @@ describe.skipIf(!canRun)('webhook-outbox-poller', () => {
       // subsequent tests still have a working queue handle.
       const fresh = getWebhookDeliveryQueue();
       expect(fresh).toBeInstanceOf(Queue);
+
+      // Wait for the rebuilt queue's ioredis client + subscriber connections
+      // to fully establish before we close them. Otherwise `afterAll`'s
+      // `__closeWebhookDeliveryQueueForTests()` races against in-flight
+      // ioredis handshake commands and the close handler rejects them with
+      // `Error: Connection is closed.` — which surfaces as an unhandled
+      // rejection that fails the whole vitest run (see Compendiq/compendiq-ce
+      // PRs #365, #373, #377). Owning the cleanup inside the test that
+      // built the queue eliminates the race entirely.
+      await fresh.waitUntilReady();
+      await __closeWebhookDeliveryQueueForTests();
     });
   });
 });
--- a/backend/src/routes/confluence/spaces.test.ts
+++ b/backend/src/routes/confluence/spaces.test.ts
@@ -25,6 +25,11 @@ vi.mock('../../core/utils/logger.js', () => ({
 
 import { spacesRoutes } from './spaces.js';
 
+// Module-scoped mocks so individual tests can re-arm `scan` to assert the
+// cross-user invalidation path (#352, finding 1).
+const mockRedisScan = vi.fn().mockResolvedValue({ cursor: '0', keys: [] });
+const mockRedisDel = vi.fn().mockResolvedValue(0);
+
 describe('Spaces routes', () => {
   let app: ReturnType<typeof Fastify>;
 
@@ -39,6 +44,12 @@ describe('Spaces routes', () => {
     app.decorate('redis', {
       get: vi.fn().mockResolvedValue(null),
       setEx: vi.fn().mockResolvedValue('OK'),
+      // RedisCache.invalidateAcrossUsers walks SCAN cursors and DELs the
+      // returned keys. Tests need both to assert the cross-user fan-out
+      // pattern; default mocks return an empty page so the SCAN walk
+      // terminates immediately.
+      scan: mockRedisScan,
+      del: mockRedisDel,
     });
 
     await app.register(spacesRoutes, { prefix: '/api' });
@@ -54,6 +65,10 @@ describe('Spaces routes', () => {
     mockQuery.mockResolvedValue({ rows: [], rowCount: 0 });
     mockGetUserAccessibleSpaces.mockResolvedValue([]);
     mockUserHasPermission.mockResolvedValue(false);
+    // Re-arm Redis SCAN to terminate immediately for tests that don't
+    // care about cache invalidation; tests that DO care override this.
+    mockRedisScan.mockReset().mockResolvedValue({ cursor: '0', keys: [] });
+    mockRedisDel.mockReset().mockResolvedValue(0);
   });
 
   it('GET /spaces includes selected spaces that are not yet synced', async () => {
@@ -249,6 +264,43 @@ describe('Spaces routes', () => {
       spaceKey: 'DEV',
       customHomePageId: null,
     });
+  });
+
+  it('PUT /spaces/:key/home invalidates the spaces cache across ALL users, not just the admin caller (#352)', async () => {
+    // Two simulated users have a cached `kb:<uid>:spaces:list` entry. The
+    // update must wipe both, otherwise non-admin viewers would keep
+    // reading the old `homepageId` for up to the spaces TTL (15 min).
+    mockUserHasPermission.mockResolvedValueOnce(true);
+    mockGetUserAccessibleSpaces.mockResolvedValueOnce(['DEV']);
+    mockQuery
+      .mockResolvedValueOnce({
+        rows: [{ space_key: 'DEV', source: 'confluence', visibility: null }],
+      })
+      .mockResolvedValueOnce({ rows: [{ space_key: 'DEV' }], rowCount: 1 });
+
+    mockRedisScan.mockResolvedValueOnce({
+      cursor: '0',
+      keys: ['kb:alice:spaces:list', 'kb:bob:spaces:list'],
+    });
+
+    const response = await app.inject({
+      method: 'PUT',
+      url: '/api/spaces/DEV/home',
+      headers: { 'content-type': 'application/json' },
+      payload: JSON.stringify({ homePageId: 42 }),
+    });
+
+    expect(response.statusCode).toBe(200);
+    // SCAN walks the shared `kb:*:spaces:*` namespace, NOT the
+    // per-admin `kb:test-user-id:spaces:*` namespace.
+    expect(mockRedisScan).toHaveBeenCalledWith('0', {
+      MATCH: 'kb:*:spaces:*',
+      COUNT: 100,
+    });
+    expect(mockRedisDel).toHaveBeenCalledWith([
+      'kb:alice:spaces:list',
+      'kb:bob:spaces:list',
+    ]);
   });
 
   it('PUT /spaces/:key/home returns 404 when the space is not in the caller\'s accessible set (#352)', async () => {
--- a/backend/src/routes/confluence/spaces.test.ts
+++ b/backend/src/routes/confluence/spaces.test.ts
@@ -1,8 +1,10 @@
 import { describe, it, expect, beforeAll, afterAll, beforeEach, vi } from 'vitest';
 import Fastify from 'fastify';
+import sensible from '@fastify/sensible';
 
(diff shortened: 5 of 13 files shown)


laboef1900 (commented):
## Automated PR Review

**PR**: #377 — feat(#352): per-space custom home page + admin/space-owner RBAC
**Branch**: feature/352-space-home -> dev
**Diff size**: 281 lines (+270/-11)
**Reviewed at**: 2026-04-27 06:50 UTC

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | backend tsc clean |
| Lint | PASS | spaces.ts clean |
| Backend Tests (scoped) | PASS | 8/8 in spaces.test.ts |
| Frontend Tests (scoped) | PASS | 53/53 in SidebarTreeView.test.tsx |
| CI: Tests | PASS | run 25020703027 |
| CI: Typecheck & Lint | PASS | run 25020703027 |
| CI: Analyze | PASS | run 25020701565 |
| CI: semgrep | PASS | scan 160170336 |
| Branch Target | PASS | feature/352-space-home -> dev |

### Docker App Test

Docker tests skipped: container shape unchanged (no compose / Dockerfile / env edits per task brief).

### E2E Browser Tests

E2E tests skipped: container shape unchanged.

### Code Review Findings

| 1 | WARNING | backend/src/routes/confluence/spaces.ts:164 | `cache.invalidate(userId, 'spaces')` only clears the **caller's** spaces cache. The `customHomePageId` field is part of every accessible user's `GET /api/spaces` response. With `TTL.spaces = 900s` (15 min, see `core/services/redis-cache.ts:434`), other users continue to see the old `homepageId` for up to 15 minutes after admin updates the home — i.e. the new home page is invisible to the rest of the org until their cache expires. Recommend either (a) using `cache.bus`/pub-sub to broadcast invalidation, or (b) a per-space identifier so all users targeting that space can be cleared, or (c) keying the spaces cache by `(userId, listVersion)` and bumping a global version on update. |
| 2 | WARNING | frontend/src/shared/components/layout/SidebarTreeView.tsx:348-359 | The "keep homepage expanded" useEffect is now dead code. After this PR the homepage node is filtered out of `tree` (lines 68-77), so adding its id to `expandedIds` has no visible effect. The accompanying comment ("keep that homepage expanded so the homepage remains clickable while its immediate children stay visible") describes the OLD behavior. Remove the effect (and the comment) to avoid future confusion. |
| 3 | WARNING | packages/contracts/src/schemas/spaces.ts:3-11 | Issue AC-4 says "Backend exposes the setting via the existing space settings contract in `@compendiq/contracts`". The `Space` schema in this file was not updated — `customHomePageId` is wire-format only. The frontend never validates it against the contract today (the file is also out of sync with the actual response shape: `key`/`name`/`pageCount` are missing), but the AC is still unmet. Add `customHomePageId: z.number().int().nullable().optional()` to keep the contract honest. |
| 4 | INFO | backend/src/routes/confluence/spaces.ts:122-126 | The redundant `getUserAccessibleSpaces` check after `userHasPermission('manage', key)` succeeded is fine as defense-in-depth, but worth a one-line comment explaining why both are needed (system_admin bypass would otherwise let an admin PUT a typo'd key and 200 instead of 404 — assuming the UPDATE matches a row). The current "Reject if the space doesn't exist or isn't accessible to the caller" comment hides the real reason. |
| 5 | INFO | backend/src/routes/confluence/spaces.ts:62-72 | The new `homepageId` resolution logic loses information: when `custom_home_page_id` is set but points to a soft-deleted page (deleted_at IS NOT NULL), the response advertises a homepageId the caller cannot fetch. Migration's `ON DELETE SET NULL` only covers hard deletes, not soft deletes via `deleted_at`. Consider `LEFT JOIN pages chp ON chp.id = cs.custom_home_page_id AND chp.deleted_at IS NULL` and treating null as "fall back to Confluence default". |
| 6 | INFO | backend/src/routes/confluence/spaces.test.ts:184-209 | The "accepts a valid in-space page" test verifies the success body but doesn't assert the UPDATE was called with the correct args — `expect(mockQuery).toHaveBeenCalledWith(expect.stringContaining('UPDATE spaces'), [42, 'DEV'])` would catch parameter-order regressions. |
| 7 | INFO | backend/src/routes/confluence/spaces.ts:152-160 | After the `getUserAccessibleSpaces(userId).includes(key)` check, the `result.rowCount === 0` branch on the UPDATE is effectively unreachable for non-admin callers (they can't be in the accessible set if the row doesn't exist). For system admins it can still fire (admin's accessible set is union of assignments + `spaces` table — they see typo'd keys as 404). Worth a comment or remove. |

### Acceptance Criteria (Issue #352)

| AC-1 | Root cause of broken "Show home content" identified and fixed | PARTIAL | Confluence spaces with custom home set work; for local spaces and Confluence spaces with no custom home / no `homepage_id`, the toggle remains a no-op (see `local-spaces.ts` — still does not select/return `homepage_id` or `custom_home_page_id`). The PR creates a path to fix it but doesn't fix it on its own |
| AC-2 | New "Home folder/page" picker UI in space settings | NOT DELIVERED | PR description: "Out of scope (deferred follow-up): The home picker UI in space settings." `frontend/src/features/spaces/SpaceSettingsPage.tsx` has no home/picker references |
| AC-3 | Setting persists; migration added; `06-data-model.md` updated | PASS | migrations/071_spaces_custom_home_page.sql + 06-data-model.md:27 |
| AC-4 | Backend exposes via `@compendiq/contracts` | PARTIAL | `customHomePageId` returned by REST but not in `packages/contracts/src/schemas/spaces.ts` — see Finding #3 |
| AC-5 | Side nav filters out the configured home entry | PASS | SidebarTreeView.tsx:68-77 + new test SidebarTreeView.test.tsx:311 |
| AC-6 | Space "Home" link uses new setting + documented fallback | PASS | spaces.ts:67-72 (COALESCE custom > Confluence > null), comment documents fallback |
| AC-7 | Permissions respected — fall back if user lacks access to home | PARTIAL | Setter rejects pinning inaccessible pages (sameSpace / sharedStandalone check), good. Viewer-side fallback when an existing home becomes RBAC-inaccessible is not implemented — `homepageId` is still surfaced and frontend `usePage(homepageId)` will 403 |
| AC-8 | Backend test for persistence + frontend test for nav | PASS | 6 new RBAC/200/400/404/null backend tests; SidebarTreeView nav test updated |
| AC-9 | Works in both themes (Graphite Honey + Honey Linen) | PASS | No theme-specific code; tree filtering is theme-agnostic |

### Story Validation (Issue #352)

**Issue intent**: Make the broken "show space home content" feature work, let admins/space-owners configure which folder/page is the home, and stop showing that page as a redundant entry in the side nav.

**Story completion**: PARTIALLY ADDRESSED

**User journey**:
- Before: Admin opens a Confluence space → toggle does nothing for unsynced/local spaces; even when it works, the home page also shows in the tree as a redundant entry. No way to override the Confluence-derived home.
- After (with this PR): Admin can call `PUT /api/spaces/:key/home` (curl/REST only — no UI) to set a custom home; the home page is hidden from the side nav and its children are promoted; `GET /api/spaces` returns the custom value. **A non-technical admin still has no UI to set the home**, so for local spaces and any Confluence space without a Confluence-derived home, the toggle still appears broken from their perspective.

**Gaps**:
1. **AC-2 explicitly deferred**: no picker UI. The user-facing fix for "the toggle does nothing" requires this UI to be usable. Without it, only operators with curl access can complete the workflow.
2. **AC-1 only partially addressed**: `local-spaces.ts` still does not select `homepage_id` / `custom_home_page_id` from the DB (PR does not touch this file). The home content view remains unreachable for local spaces unless the Confluence path is used (which doesn't apply to local spaces by definition).
3. **AC-4 partial**: contract schema not updated.
4. **AC-7 partial**: no viewer-side fallback if the configured home becomes inaccessible.
5. **Cache invalidation is per-user only** — see Finding #1. After admin updates, every other user keeps the stale home for up to 15 minutes.

**Evidence**: Backend changes (spaces.ts:98-170, migration 071), frontend tree filter (SidebarTreeView.tsx:68-77), 8 backend tests, 1 updated frontend test, ERD updated. PR explicitly self-acknowledges the picker UI gap.

### Suggested Actions

1. **[SHOULD FIX]** Cache invalidation across users — Finding #1 is a real correctness bug visible to end users. Pick a strategy (cache-bus broadcast, per-space cache key, or version stamp) before merge or open a follow-up issue with high priority.
2. **[SHOULD FIX]** Remove the now-dead "keep homepage expanded" useEffect at SidebarTreeView.tsx:348-359 along with its outdated comment — small cleanup that prevents future contributors from rebuilding around stale assumptions.
3. **[CONSIDER]** Add `customHomePageId` to `packages/contracts/src/schemas/spaces.ts` to satisfy AC-4 and keep the contract in sync with the route response.
4. **[CONSIDER]** Filter out soft-deleted custom home pages in the GET join (Finding #5) — `ON DELETE SET NULL` only covers hard deletes, but the codebase soft-deletes via `deleted_at`.
5. **[CONSIDER]** Open a follow-up issue tracking the deferred picker UI and the local-spaces homepage exposure so AC-1/AC-2 don't drift.
6. **[OPTIONAL]** Tighten the success test to assert UPDATE arg shape (Finding #6).

### Verdict: COMMENT

**Reason**: Backend RBAC and migration are sound; tests and CI are green; the change is small and reversible. However: (a) cache invalidation is per-user with a 15-minute TTL, so other users see stale `homepageId` after an admin update — a real, visible regression vector; (b) story is explicitly partially addressed (no picker UI; local-spaces home read-path untouched); (c) contracts schema not updated; (d) one block of dead code in SidebarTreeView. None of these are merge-blockers if the cache invalidation is acknowledged and tracked, but they prevent an APPROVE.

---
Reviewed by PR Reviewer Agent | 2026-04-27

laboef1900:
Pushed `70e9c7c` addressing the three reviewer findings + INFO #2.

### Finding 1 — Stale spaces cache for non-admin users (WARNING)

`PUT /api/spaces/:key/home` now calls `cache.invalidateAcrossUsers('spaces')` instead of the per-user `cache.invalidate(userId, 'spaces')`. Implemented as a new `RedisCache.invalidateAcrossUsers(type)` method that walks the SCAN cursor for `kb:*:{type}:*` against the shared Redis. This works in both single-pod and multi-pod deployments without needing pub/sub (Redis is the shared store). Considered the existing in-process `cache-bus` pattern but it's designed for in-process module-level cached values that need cross-pod re-load broadcasts — not Redis-key invalidation, where SCAN+DEL is sufficient.

Regression test in `backend/src/routes/confluence/spaces.test.ts` asserts SCAN is called with `MATCH: 'kb:*:spaces:*'` and DEL receives both simulated users' keys.

### Finding 2 — AC-1 still broken for local spaces

Extended the SELECT in `backend/src/routes/knowledge/local-spaces.ts` to include `cs.custom_home_page_id`, and mapped both `homepageId` (stringified, matches Confluence wire format) and `customHomePageId` into the response. Local spaces have no Confluence-derived homepage, so the resolution collapses to "custom_home_page_id or null". Added two tests: one for the override-set path, plus a SELECT regression guard so the column can never be silently dropped from the query again.

### Finding 3 — AC-4 contracts not updated

Added `customHomePageId: z.number().nullable().optional()` to `SpaceSchema` in `packages/contracts/src/schemas/spaces.ts`. Optional + nullable mirrors the wire reality: present on Confluence spaces and (now) local spaces; null when no override is set. Also added the field to the inline `Space` interface in `frontend/src/shared/hooks/use-spaces.ts` for symmetry.

### INFO #2 — Dead useEffect in SidebarTreeView

Removed the `useEffect` at lines 348-359 that pinned the homepage as expanded. Lines 68-77 of `buildTree` now filter the homepage out of the tree, so the effect could never find its target node. Comment was misleading too. Existing `#352` test (`hides the homepage from the tree but promotes its children + other roots`) still passes.

### Verification

```
backend/src/core/services/redis-cache.test.ts        61 passed
backend/src/routes/confluence/spaces.test.ts          9 passed
backend/src/routes/knowledge/local-spaces.test.ts    17 passed
frontend/src/shared/components/layout/
    SidebarTreeView.test.tsx                         53 passed

npm run typecheck                                    clean
npm run lint                                         clean
```

### Architecture diagrams

Per CLAUDE.md rule 6: `docs/architecture/06-data-model.md` already documents the `pages ||--o{ spaces : "is custom home of (#352)"` relationship at line 27. The doc has no standalone `spaces` entity definition (pre-existing gap, not introduced by this PR), so no inline column update is needed.

laboef1900 (commented):
## Automated PR Re-Review

**PR**: #377 -- feat(#352): per-space custom home page + admin/space-owner RBAC
**Branch**: feature/352-space-home -> dev
**HEAD**: 70e9c7c (fix commit on top of 3c9ee46)
**Diff size**: 497 lines (+472/-25)
**Reviewed at**: 2026-04-28 08:15 UTC

This is a re-review of the fix commit `70e9c7c` that addresses three findings from the prior automated review (WARNING + 2 INFO + 1 dead-code). Validation focuses on the deltas, not the original PR.

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | Full repo typecheck clean (`npm run typecheck`) |
| Lint | PASS | Backend + frontend ESLint clean (no new warnings) |
| Backend Tests (scoped) | PASS | redis-cache 61/61, spaces 9/9, local-spaces 17/17 |
| Frontend Tests (scoped) | PASS | SidebarTreeView 53/53 |
| CI: Tests | FAIL | 2502 tests pass, 1 unhandled rejection in `webhook-outbox-poller.test.ts` (Redis "Connection is closed" on async teardown). **Not caused by this PR — webhook-outbox-poller files are untouched.** Earlier commit `3c9ee46` on the same branch passed CI fully; the flake is environment-specific. |
| CI: Typecheck & Lint | PASS | |
| CI: CodeQL | PASS | |
| CI: Semgrep | PASS | |
| Branch Target | PASS | feature/352-space-home -> dev |

### Code Review Findings (deltas only)

| 1 | INFO | backend/src/core/services/redis-cache.ts:506-512 | SCAN+DEL fan-out is the correct production-safe pattern (verified vs Redis docs: `redis.io/commands/scan` — non-blocking by design, COUNT=100 hint, MATCH applied post-retrieval). At current scale (likely <100K total `kb:*` keys) the call is sub-second. **Watch-out for future scale**: if total `kb:*` cardinality grows past 1M (e.g. heavy LLM cache reuse with TTL=3600), the cross-user invalidation will return many empty SCAN pages because MATCH filters server-side after the cursor walk. Consider re-evaluating if Redis CPU on this call ever shows up in a flame graph. Not actionable now. |
| 2 | INFO | backend/src/routes/confluence/spaces.ts:80-90 | The `unsyncedSelections` branch of GET /api/spaces builds objects with `homepageId: null` but does not set `customHomePageId`. Since `SpaceSchema.customHomePageId` is `.optional()`, this validates fine, but it's a small inconsistency vs the synced branch. Suggest mirroring with an explicit `customHomePageId: null` for symmetry. Cosmetic. |
| 3 | INFO | backend/src/core/services/redis-cache.ts:506-512 | A SCAN-during-write race exists: a `GET /api/spaces` request that completes its read after the SCAN cursor finishes but before its own SET fires would re-cache stale data, surviving until next 15-min TTL. This is structurally inherent to any SCAN+DEL pattern and is acceptable here (mitigated by short TTL + next admin update re-runs SCAN). Author's choice of SCAN+DEL over cache-bus pub/sub is sound: Redis IS the shared store, so DEL is sufficient — no cross-pod coordination needed. Documented in the JSDoc. |

### Resolution of Prior Findings

| # | Prior Finding | Status | Evidence |
|---|--------------|--------|----------|
| W-1 | Stale spaces cache for non-admins (`cache.invalidate(userId, 'spaces')` only cleared the calling admin) | RESOLVED | `redis-cache.ts:506-512` adds `invalidateAcrossUsers(type)` walking `kb:*:{type}:*`; `spaces.ts:167` calls it instead of per-user invalidate. New test in `spaces.test.ts:269-301` asserts SCAN MATCH=`kb:*:spaces:*` with multiple-user keys and verifies DEL receives both `kb:alice:spaces:list` and `kb:bob:spaces:list`. |
| W-2 | AC-1 broken for local spaces (SELECT did not include `homepage_id`/`custom_home_page_id`) | RESOLVED | `local-spaces.ts:65-80` extends SELECT with `cs.custom_home_page_id`; mapper at lines 105-107 emits `homepageId` (stringified) and `customHomePageId` (number). Wire format matches the Confluence-spaces route exactly. New test in `local-spaces.test.ts:114-139` covers the override path; new regression guard at lines 141-150 asserts the SELECT clause contains `cs.custom_home_page_id`. |
| W-3 | Contracts schema not updated — `customHomePageId` was wire-only | RESOLVED | `packages/contracts/src/schemas/spaces.ts:9-16` adds `customHomePageId: z.number().nullable().optional()` to `SpaceSchema`. JSDoc explains the resolution semantics (custom override → fallback to Confluence default). |
| INFO-1 | Dead `useEffect` at `SidebarTreeView.tsx:348-359` pinning a homepage that's now filtered out of the tree | RESOLVED | Effect block removed cleanly. Verified at `SidebarTreeView.tsx:39-80` that `buildTree` filters the homepage from `roots` (line 72) and promotes its children (line 71); the `homepageId` is no longer present in `nodeMap`-mapped roots, so the prior effect could never have set it as expanded. No callers depended on the effect's side-effect. |

### Acceptance Criteria (Issue #352)

| AC-1 | "Show home content" toggle works for both Confluence and local spaces | PASS | Confluence: `spaces.ts:67-72` (existing) + local: `local-spaces.ts:105-107` (new). Both surface `homepageId` as wire-format string. |
| AC-2 | Admin / space-owner can override Confluence-derived homepage | PASS | PUT `/api/spaces/:key/home` gated on `userHasPermission('manage', key)` (admin via early-return + space_admin role per migration 039). |
| AC-3 | Custom home is hidden from sidebar tree, children promoted | PASS | `SidebarTreeView.tsx:68-77` filters and promotes. 53/53 tests cover edge cases (homepage with grandchildren, multiple roots, missing homepage). |
| AC-4 | Wire contract typed in `@compendiq/contracts` | PASS | `SpaceSchema.customHomePageId` added with documented semantics. |

### Story Validation

**Issue intent**: Make the per-space "home page" both settable (admin override) and visually de-duplicated (don't show home as a sidebar entry when it's already reachable via the Home link).
**Story completion**: PARTIALLY ADDRESSED — same as before. The "settable home" backend + data layer + RBAC shipped in this PR (with cross-user cache fanout now correct). The home-picker UI in space settings is **explicitly deferred** in the PR description and remains an open follow-up tracked under #352.
**User journey**: Admin calls `PUT /api/spaces/MYKEY/home` with `{homePageId: N}` → all users (not just the admin) see the new home on next read because `invalidateAcrossUsers('spaces')` wipes every `kb:*:spaces:*` key. Tree view drops the homepage entry; children promote. Local-space users now see `homepageId`/`customHomePageId` in the GET response so the existing "Show home content" toggle works.
**Gaps**: Home-picker UI deferred (acknowledged in PR body and noted by author).
**Evidence**: Verified test files cover the regressions (cross-user fan-out, local-spaces SELECT, schema typing).

### Scaling Concerns (per request)

- **SCAN+DEL with COUNT=100**: production-safe per Redis docs. Non-blocking, cursor-based. Matches existing `clearAttachmentFailures` pattern in the codebase.
- **MATCH applied server-side after retrieval**: at current scale, fine. At 10x growth in non-spaces `kb:*` keys, the SCAN walk visits all of them and returns mostly empty pages. Theoretically O(N) on Redis main thread but each chunk is microsecond-scale. INFO-only.
- **Race condition during SCAN**: a concurrent `cache.set` interleaved with the SCAN walk could re-populate a stale entry that survives until next TTL (15 min). Mitigated by next admin update + short TTL. Author's design rationale (Redis-as-store, not cache-bus) is correct because there's no cross-pod in-process state to broadcast.
- **No Redis Cluster hash-tag**: pattern doesn't pin to a slot. Acceptable for current single-node Redis deployment.

### Suggested Actions

1. **[OPTIONAL]** Add explicit `customHomePageId: null` to the `unsyncedSelections` mapper in `confluence/spaces.ts:80-90` for symmetry with the synced-spaces branch. Cosmetic; current schema accepts the omission.
2. **[OPTIONAL]** When/if `kb:*` cardinality grows past ~1M total keys (likely needs heavy LLM cache reuse), revisit SCAN-MATCH performance. Until then, no action.

### Verdict: APPROVE

**Reason**: All three prior findings (W-1 stale cache, W-2 AC-1 broken for local spaces, W-3 contracts) are resolved with targeted fixes, regression tests, and clear JSDoc. Dead `useEffect` removed. Static checks clean. The CI test failure is an unrelated pre-existing flake in `webhook-outbox-poller.test.ts` (Redis async teardown), in code untouched by this PR. Author's design choice of SCAN+DEL over cache-bus is sound and matches Redis docs guidance. New INFO findings are scaling watch-outs, not blockers.

---

laboef1900:
Cherry-picked the webhook-outbox-poller test-flake fix from #381 onto this branch to unblock CI. Root cause was a race in `webhook-outbox-poller.test.ts` where the `idempotent teardown` test rebuilt a BullMQ Queue (2 ioredis connections) at the end of the test body, then `afterAll` closed it before the second connection finished its handshake → `Error: Connection is closed` unhandled rejection failed the run. The fix awaits `Queue#waitUntilReady()` then explicitly closes inside the test scope so `afterAll` finds `deliveryQueue === null`. New CI run will validate. (Same root-cause fix lives in #381 against `dev` and will dedupe at merge.)

PR #382, opened 2026-04-28
fix(llm-providers): allowlist private-network base URLs before SSRF check

laboef1900:
## Why

The Compendiq EE stack lets admins point Compendiq at any OpenAI-compatible LLM endpoint via Settings → LLM. But trying to add an LM Studio at `http://192.168.178.185:1234/v1` (or an on-prem Ollama at `http://10.0.0.5:11434/v1`, etc.) returned a generic "Request failed" — actually a `400` from the SSRF guard rejecting the RFC-1918 IP **before** the URL had a chance to enter the allowlist.

The Confluence test-connection flow already had the right ordering (`addAllowedBaseUrl` → `validateUrl` → on failure `removeAllowedBaseUrl`) for exactly this reason — see `backend/src/routes/foundation/settings.ts:236-251`. This PR mirrors that pattern in `llm-providers.ts`.

## What changes

`backend/src/routes/llm/llm-providers.ts`:
- POST: `addAllowedBaseUrl(input.baseUrl)` → `validateUrl(input.baseUrl)` (sync; protocol check still runs first) → `createProvider(...)`. On any failure the speculative allowlist entry is revoked with `removeAllowedBaseUrl` so we never leave a dangling allowlist over a non-existent provider.
- PATCH: same pattern, only when `patch.baseUrl` is being changed; the entry is revoked on validate failure, on `updateProvider` throw, and on 404.

The runtime path is unchanged: `llm-provider-bootstrap.ts:73` already calls `addAllowedBaseUrlSilent` for every existing row at startup, so private-network providers worked once a row existed (the issue was *only* the create/update path).

## Defence-in-depth notes

- `LlmProviderInputSchema` enforces `http(s)` at Zod time (`packages/contracts/src/llm.ts:13`), so `file:/javascript:/data:` URLs never reach the SSRF guard.
- `validateUrl` runs the protocol check **before** the allowlist short-circuit (`ssrf-guard.ts:238` runs before line 243), so even an allowlisted non-HTTP origin would be rejected.
- The user submitting the URL is an authenticated admin; the same trust model as the existing Confluence private-network allowance.

## Tests

The 2 original "loopback returns 400" tests are renamed and inverted to assert the new "loopback / RFC-1918 / public-to-RFC-1918 transition is accepted" behaviour — the correctness bar this PR pins. A new test for duplicate-name collision proves the allowlist is revoked when the DB row never lands.

```
SSRF guard — private-network base URLs accepted ...
  ✓ POST with loopback baseUrl is now accepted and allowlisted
  ✓ POST with RFC-1918 baseUrl is accepted and allowlisted
  ✓ PATCH from public to RFC-1918 baseUrl is accepted
  ✓ POST with duplicate name revokes the speculative allowlist entry
```

## Repro

Pre-fix: in Settings → LLM, "Add provider" with base URL `http://192.168.178.185:1234/v1`, auth `none` → toast "Request failed". Backend logs: `400 SSRF blocked: cannot connect to internal/private network`.

Post-fix: same flow → 201, provider listed, allowlist hit on first chat completion.

## Risks

- Loopback URLs (`http://127.0.0.1/...`) are now also accepted at create-time. This matches Confluence settings behavior. The user is an authenticated admin; the trust gate is the admin role plus the deliberate URL submission.

### DIFF
--- a/backend/src/routes/llm/llm-providers.test.ts
+++ b/backend/src/routes/llm/llm-providers.test.ts
@@ -268,6 +268,74 @@ describe.skipIf(!dbAvailable)('SSRF guard — private-network base URLs accepted
     // The shared origin is still owned by Shared-A — must remain allowlisted.
     expect(() => validateUrl('http://10.0.0.70/v1')).not.toThrow();
   });
+
+  // ──────────────────────────────────────────────────────────────────────
+  // INFO #5: PATCH revoke-on-failure paths now have explicit coverage.
+  // The route does `await revokeAllowlistIfUnused(patch.baseUrl)` in three
+  // branches (catch-on-validate, catch-on-updateProvider, 404-not-found);
+  // only the validate branch was implicitly covered by inspection of POST.
+  // The two tests below pin the remaining branches so a regression that
+  // dropped the revoke from either branch would be caught here.
+  // ──────────────────────────────────────────────────────────────────────
+
+  it('PATCH against a non-existent provider id returns 404 and revokes the speculative allowlist entry', async () => {
+    const { validateUrl, SsrfError } = await import('../../core/utils/ssrf-guard.js');
+
+    // No DB row exists at this UUID — `updateProvider` returns undefined,
+    // taking the `if (!updated) … 404` branch.
+    const r = await app.inject({
+      method: 'PATCH',
+      url: '/api/admin/llm-providers/00000000-0000-4000-8000-000000000000',
+      headers: { authorization: `Bearer ${adminToken}`, 'content-type': 'application/json' },
+      payload: JSON.stringify({ baseUrl: 'http://10.0.0.99/v1' }),
+    });
+    expect(r.statusCode).toBe(404);
+
+    // The 404 branch must call `revokeAllowlistIfUnused(patch.baseUrl)`.
+    // Nothing else references this origin, so the allowlist entry should
+    // be revoked and `validateUrl` should fall back to the SSRF block.
+    expect(() => validateUrl('http://10.0.0.99/v1')).toThrow(SsrfError);
+  });
+
+  it('PATCH name collision (updateProvider throws) revokes the speculative allowlist entry', async () => {
+    const { validateUrl, SsrfError } = await import('../../core/utils/ssrf-guard.js');
+
+    // Seed two distinct providers with distinct names + non-overlapping origins.
+    const a = await app.inject({
+      method: 'POST', url: '/api/admin/llm-providers',
+      headers: { authorization: `Bearer ${adminToken}`, 'content-type': 'application/json' },
+      payload: JSON.stringify({ name: 'Throw-A', baseUrl: 'http://10.0.0.80/v1', authType: 'none', verifySsl: true }),
+    });
+    expect(a.statusCode).toBe(201);
+    const b = await app.inject({
+      method: 'POST', url: '/api/admin/llm-providers',
+      headers: { authorization: `Bearer ${adminToken}`, 'content-type': 'application/json' },
+      payload: JSON.stringify({ name: 'Throw-B', baseUrl: 'http://10.0.0.81/v1', authType: 'none', verifySsl: true }),
+    });
+    expect(b.statusCode).toBe(201);
+
+    // PATCH B's name to collide with A's, and at the same time bring in a
+    // brand-new baseUrl (origin http://10.0.0.82) so the speculative
+    // allowlist entry is observable. `updateProvider` will throw on the
+    // unique-name constraint — taking the catch-on-updateProvider branch.
+    const fail = await app.inject({
+      method: 'PATCH',
+      url: `/api/admin/llm-providers/${b.json().id}`,
+      headers: { authorization: `Bearer ${adminToken}`, 'content-type': 'application/json' },
+      payload: JSON.stringify({ name: 'Throw-A', baseUrl: 'http://10.0.0.82/v1' }),
+    });
+    expect(fail.statusCode).toBeGreaterThanOrEqual(400);
+
+    // The catch branch must revoke the speculative allowlist entry —
+    // nothing references the new origin, so it should be revoked.
+    expect(() => validateUrl('http://10.0.0.82/v1')).toThrow(SsrfError);
+    // Provider A's original origin must be untouched.
+    expect(() => validateUrl('http://10.0.0.80/v1')).not.toThrow();
+    // Provider B's original origin must also be untouched (the throw
+    // happens AFTER `previous` is captured but BEFORE the GC-on-success
+    // path runs, so the previous origin stays allowlisted).
+    expect(() => validateUrl('http://10.0.0.81/v1')).not.toThrow();
+  });
 });
 
 describe.skipIf(!dbAvailable)('DELETE race conditions return 409 (not 500)', () => {
--- a/backend/src/routes/llm/llm-providers.test.ts
+++ b/backend/src/routes/llm/llm-providers.test.ts
@@ -149,7 +149,9 @@ describe.skipIf(!dbAvailable)('SSRF guard — private-network base URLs accepted
     expect(r.json().baseUrl).toBe('http://10.0.0.5:1234/v1');
   });
 
-  it('POST with duplicate name revokes the speculative allowlist entry', async () => {
+  it('POST with duplicate name revokes the speculative allowlist entry (verified directly)', async () => {
+    const { validateUrl, SsrfError } = await import('../../core/utils/ssrf-guard.js');
+
     // Seed the row that will collide.
     await app.inject({
       method: 'POST', url: '/api/admin/llm-providers',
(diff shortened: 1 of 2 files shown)


laboef1900 (commented):
## Automated PR Review

**PR**: #382 — fix(llm-providers): allowlist private-network base URLs before SSRF check
**Branch**: `fix/llm-providers-ssrf-allowlist-order` → `dev`
**Diff size**: ~113 lines (+93/-20), 2 files
**Reviewed at**: 2026-04-27 12:30 UTC

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | `tsc -p backend/tsconfig.json --noEmit` clean |
| Lint | PASS | `eslint` clean on changed files |
| Targeted tests | PASS | `npm test -w backend src/routes/llm/llm-providers.test.ts` → 11/11 pass |
| SSRF guard tests | PASS | `npm test -w backend src/core/utils/ssrf-guard.test.ts` → 61/61 pass |
| Confluence settings tests | PASS | `npm test -w backend src/routes/foundation/settings.test.ts` → 26/26 pass (no regression in the parity flow) |
| CI: Tests | PASS | upstream CI green |
| CI: Typecheck & Lint | PASS | upstream CI green |
| CI: CodeQL | PASS | no new alerts |
| CI: Semgrep | PASS | no new findings |
| Branch Target | PASS | `feature/* → dev` ✓ |

### Docker App Test

Skipped — not part of this review's scope (dev stack is in use; the change is a route-level fix and unit-tested end-to-end with PG running on :5433).

### E2E Browser Tests

Skipped — no admin LLM-provider Playwright spec in the repo, and the change is server-side only.

### Code Review Findings

The fix is small, the pattern matches the prior Confluence work on `settings.ts:236-251`, the new tests pin the new contract, and CI is green. I went hunting for the security-sensitive failure modes the user flagged. Findings below.

| 1 | WARNING | `backend/src/routes/llm/llm-providers.ts:71-76` | **Concurrent-failure revoke race.** Two POSTs against the same `baseUrl` (e.g. one with a duplicate name) race: A's `addAllowedBaseUrl` adds the origin, B's `addAllowedBaseUrl` is a no-op (set semantics), A's `createProvider` succeeds, B's `createProvider` throws unique-violation, B's catch-block calls `removeAllowedBaseUrl(input.baseUrl)` — which removes the origin A's now-persisted row legitimately depends on. The provider row stays in DB but every subsequent runtime call to it gets SSRF-blocked until the next pod restart re-bootstraps the allowlist. Same shape applies to PATCH paths at lines 110-117. The window is small (single INSERT) and recovery is `bootstrapLlmProviders` on restart, but with the multi-pod pub/sub bus (`SSRF_ALLOWLIST_CHANNEL`) the revoke propagates to peer pods and amplifies the impact. A safer revoke would be conditional on "no provider currently uses this origin" — i.e. revoke only when `SELECT 1 FROM llm_providers WHERE base_url=$1` is empty. |
| 2 | WARNING | `backend/src/routes/llm/llm-providers.ts:108-114` | **PATCH leaks the previous baseUrl in the allowlist.** When PATCH changes baseUrl from URL_A to URL_B, the new origin is allowlisted but the OLD origin is never revoked. Compare to `settings.ts:108-110` which explicitly does `if (oldUrl && oldUrl !== body.confluenceUrl) removeAllowedBaseUrl(oldUrl)` for exactly this. Per-pod allowlist accumulates stale URLs over time as admins re-target providers. The stale URL was admin-trusted at one point but the admin has since revoked it from the active config. Bootstrap on next restart heals it (it re-reads `llm_providers.base_url` and won't see the old one), but for the lifetime of a long-running pod these accumulate. Fix: read the current `base_url` for `id` before the UPDATE, and after success call `removeAllowedBaseUrl(oldBaseUrl)` if it differs from the new one (and no other provider uses it — see #1). |
| 3 | WARNING | `backend/src/routes/llm/llm-providers.test.ts:152-174` | **The "duplicate name revokes the speculative allowlist entry" test does not actually verify the revoke.** It asserts `count(*) = 0` for the rejected base_url — proving "no row was created", which is true regardless of the revoke logic. The revoke claim in the comment is unverified. Add an assertion against `getAllowedBaseUrlCount()` or `clearAllowedBaseUrls()` + spy on `removeAllowedBaseUrl` to actually pin the security invariant. As written, removing the `removeAllowedBaseUrl` calls from the route would still pass this test. |
| 4 | INFO | `backend/src/routes/llm/llm-providers.ts:60` (and :99) | **DNS rebinding mitigation is no longer run at create-time.** `assertNonSsrfUrl` (which calls `validateUrlWithDns` → `resolveAndValidateIp`) was the only place the user-supplied URL got DNS-resolved before being trusted. The PR replaces it with sync `validateUrl` because the allowlist short-circuits before any DNS work anyway. Net effect: an admin pasting a domain whose A-record currently points at `169.254.169.254` (or any private IP) is no longer rejected at create time. The threat model is still "trusted authenticated admin", and once allowlisted the URL was already exempt from DNS validation at runtime, so this is a small reduction in defence-in-depth, not an exploit. The PR description is right that the protocol gate stays. Worth noting in a code comment so future readers don't assume DNS-rebinding is blocked here. |
| 5 | INFO | `backend/src/routes/llm/llm-providers.test.ts:135-150` | **PATCH revoke-on-failure path is not tested.** Three branches do `removeAllowedBaseUrl(patch.baseUrl)`: catch-on-validate, catch-on-`updateProvider`, and the 404-not-found branch. Only the validate path is implicitly covered (and only by inspection of POST). A test that PATCHes a non-existent UUID with `baseUrl=http://10.0.0.99/v1` would pin the 404 branch. A test that PATCHes id=A with a name that collides with id=B's name would pin the throw branch. |
| 6 | INFO | `backend/src/routes/llm/llm-providers.ts:68,108` | **`let provider;` / `let updated;` rely on flow-typing.** Works because both `try` blocks definitely assign before use and the catch always re-throws, so TS narrows `undefined` away after the try. `tsc --strict --noUncheckedIndexedAccess` is happy. Just calling out: `const provider = await (async () => { try { return await createProvider(input); } catch (err) { removeAllowedBaseUrl(input.baseUrl); throw err; } })();` would make the same shape readable as a single expression. Style only — no need to change. |
| 7 | INFO | `backend/src/routes/llm/llm-providers.ts:77-81` | **Failed creates are not audit-logged.** A failing `createProvider` (e.g. unique-name violation) throws past `emitLlmAudit`, so neither the create-attempt nor the speculative-then-revoked allowlist write makes it into the audit trail. Pre-existing gap (not introduced by this PR), but the speculative-allowlist pattern adds a new class of "allowlist mutated then unmutated" events that would be useful for forensic visibility on a security-sensitive surface. Consider an `llm_provider_create_failed` admin-audit event with `metadata: { baseUrl, reason }` — symmetric with the success log. |
| 8 | INFO | `backend/src/routes/llm/llm-providers.ts:53-57` | **The doc comment is correct and load-bearing — keep it.** The "Protocol restrictions still apply post-allowlist because validateUrl runs the protocol check before short-circuiting" line in the new comment is the actual security argument for why this re-ordering is safe. Worth adding a one-line link to `ssrf-guard.ts:238` (the protocol-check) and `:243` (the allowlist short-circuit) so a future refactor that swaps the order of those two checks in `validateUrl` knows it's a load-bearing invariant for this caller. |

### Story Validation

No linked issue, but the PR body is unusually thorough and self-justifying — describing the user-visible bug ("LM Studio at `http://192.168.178.185:1234/v1` returns 'Request failed'"), the root cause (`assertNonSsrfUrl` → `addAllowedBaseUrl` ordering), the prior-art reference (Confluence flow at `settings.ts:236-251`), and the defence-in-depth argument. This is essentially a "fix admin UX while preserving the security invariant" change.

**User journey**: Before — admin pastes private-network URL, gets generic "Request failed" toast (actually a 400 from SSRF guard). After — admin pastes private-network URL → 201, provider listed, allowlist hit on first chat completion. Direct fix for the symptom.

**Story completion**: FULLY ADDRESSED for the create/update path. The PR mirrors the Confluence pattern accurately. The remaining concerns (#1, #2, #3 above) are about the *robustness* of the new pattern, not whether the fix solves the original user problem — it does.

### Suggested Actions

1. **[SHOULD FIX]** #1 — Make the failure-revoke conditional on "no other provider uses this origin" (or guard the revoke under a transaction that includes the SELECT). This addresses both the concurrent-POST race and the partial overlap with #2.
2. **[SHOULD FIX]** #2 — On successful PATCH that changes baseUrl, revoke the OLD origin (with the same "no other provider" guard). Mirrors `settings.ts:108-110`.
3. **[SHOULD FIX]** #3 — Strengthen the "duplicate name revokes" test to assert the allowlist size actually dropped (or spy on `removeAllowedBaseUrl`). The current test passes even if the revoke logic is removed.
4. **[CONSIDER]** #5 — Add tests for the PATCH 404 branch and the `updateProvider` throw branch.
5. **[CONSIDER]** #4 — One-line code comment noting DNS-rebinding is intentionally NOT re-validated at create-time and why (admin-trust gate + once-allowlisted-always-skipped semantics of the existing `validateUrlWithDns`).
6. **[CONSIDER]** #7 — Symmetric `llm_provider_create_failed` admin-audit event.

### Verdict: COMMENT

**Reason**: The fix is correct for the stated user problem, follows the existing Confluence pattern, all CI is green, and all the tests I ran locally pass. But the speculative-allowlist pattern has two real correctness issues — concurrent failure can revoke another caller's legitimate entry (#1), and PATCH leaks the previous baseUrl on every URL change (#2) — and the "duplicate name revokes" test doesn't actually verify the revoke (#3). None of these are exploitable by a non-admin (the route is admin-gated), and all are recoverable on pod restart, but they're worth fixing before merge or in a fast-follow. The DNS-rebinding regression at create-time (#4) is a minor reduction in defence-in-depth that's acceptable under the admin-trust model.

---
Reviewed by PR Reviewer Agent | 2026-04-27

laboef1900:
Addressed all three review findings in the latest commit.

**Finding 1 — Concurrent revoke race**: introduced `revokeAllowlistIfUnused(baseUrl)` that queries `llm_providers` for any surviving row at the same origin (origin-keyed match via case-insensitive `SUBSTRING` regex on `base_url` so path / scheme casing differences don't escape it). All revoke paths in POST and PATCH now route through it.

**Finding 2 — PATCH baseUrl change leaked old origin**: PATCH now captures the previous `baseUrl` via `getProviderById` before mutation; after a successful `baseUrl` change, `revokeAllowlistIfUnused` runs against the old origin so it doesn't accumulate.

**Finding 3 — Test gap on revoke verification**: the duplicate-name test now asserts the allowlist state directly via the exported `SsrfError` + `validateUrl`. Plus three new tests:
- Concurrent sibling protection (failed POST with shared origin must not strip the successful sibling's allowlist)
- PATCH GC happy path (single owner of an origin migrates → old origin revoked)
- PATCH GC negative (two providers share an origin, one migrates → shared origin remains allowlisted for the other)

DNS-rebinding-at-create stays a known design property as discussed.

laboef1900:
Addressed the actionable INFO findings from the review in `53bccef`:

**INFO #5 — PATCH revoke-on-failure paths now have explicit coverage.**
Two new tests pin the previously-implicit branches:

- *PATCH against a non-existent provider id returns 404 and revokes the speculative allowlist entry.* PATCHes `00000000-0000-4000-8000-000000000000` with `baseUrl=http://10.0.0.99/v1`; asserts 404 + `validateUrl(...)` throws `SsrfError` after the request (the 404 branch revoked the speculative entry).
- *PATCH name collision (updateProvider throws) revokes the speculative allowlist entry.* Seeds `Throw-A` and `Throw-B`, then PATCHes B's name to collide with A while introducing a brand-new origin; asserts 4xx + the new origin's speculative allowlist entry is revoked, while A's and B's *original* origins stay allowlisted (the throw happens before the GC-on-success path runs, so the previous origin isn't dropped).

16/16 tests pass with the new coverage; the 14/14 baseline becomes 16/16.

**INFO #4 + #8 — DNS rebinding + SSRF guard ordering doc comment.** The `addAllowedBaseUrl(input.baseUrl)` site now carries a multi-line comment explaining (a) the load-bearing protocol-then-allowlist ordering inside `validateUrl` (so a future `ssrf-guard.ts` refactor that swaps those two checks knows it'll silently break this caller's defence-in-depth), and (b) explicitly that DNS rebinding at create-time is intentionally NOT validated here (`validateUrl` vs `assertNonSsrfUrl` choice), under the authenticated-admin trust model — so a future reader doesn't assume DNS-rebinding is blocked.

Skipped the remaining INFO items as noted in the original review:
- **#6** is style-only ("no need to change").
- **#7** is a pre-existing audit-logging gap not introduced by this PR; better as a separate change.

`tsc --noEmit` clean; `eslint` clean on the changed files.

laboef1900 (commented):
## Automated PR Review (Round 3)

**PR**: #382 — fix(llm-providers): allowlist private-network base URLs before SSRF check
**Branch**: fix/llm-providers-ssrf-allowlist-order -> dev
**Diff size**: 351 lines (+331/-20)
**HEAD**: 53bccef
**Reviewed at**: 2026-04-28

This is a re-review of commit `53bccef` ("test(llm-providers): pin remaining PATCH revoke-on-failure branches") on top of the previously-reviewed `c49edb2`. The actionable INFO items from rounds 1–2 are now addressed.

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | tsc --noEmit clean across @compendiq/contracts, backend, frontend |
| Lint | PASS | eslint --max-warnings=0 clean (only the unrelated boundaries v6-migration notice) |
| Backend Tests | SKIPPED | dbAvailable=false on review host (port 5433 owned by user dev stack); 16/16 it() blocks present and run on CI |
| Frontend Tests | N/A | This PR only touches backend |
| Production Build | N/A | Not run — only backend route + test changed |
| npm audit | INFO | 4 moderate, 0 high, 0 critical (unchanged from base) |
| Branch Target | PASS | feature/* -> dev as required |

### Docker App Test

| Check | Status |
|-------|--------|

Docker tests skipped: port 5433 held by user's dev stack (consistent with #383/#384/#385 reviews on the same host).

### E2E Browser Tests

| Project | Status |
|---------|--------|
| auth | SKIPPED |
| chromium | SKIPPED |

E2E tests skipped: backend-only change behind admin auth; no UI surface affected.

### Code Review Findings

| 1 | INFO | backend/src/routes/llm/llm-providers.ts:101-112 | Doc comment says runtime is "exempt from DNS validation by design"; technically the runtime call path in `openai-compatible-client.ts` is exempt from ALL SSRF validation (it calls `undiciFetch` directly with `cfg.baseUrl` and never invokes `validateUrl`/`assertNonSsrfUrl`). Phrasing is defensible under the trust model but a curious reader might assume `isBlockedIp` still runs at request time. Optional wording tweak. |

No CRITICAL or WARNING findings.

### Round-3 Specific Verifications

**1. Do the new tests genuinely pin their branches?**

*Test "PATCH against a non-existent provider id returns 404 and revokes the speculative allowlist entry"* (test.ts:281–298):
- Trace: PATCH valid-but-nonexistent UUID with `baseUrl=http://10.0.0.99/v1`.
- Route flow: `getProviderById` → null; `addAllowedBaseUrl` adds 10.0.0.99; `validateUrl` short-circuits on the just-added allowlist entry; `updateProvider` runs `UPDATE … WHERE id=$1 RETURNING *` and returns 0 rows → `updated` is null → enters `if (!updated)` at line 179 → `revokeAllowlistIfUnused('http://10.0.0.99/v1')` runs (DB count = 0 → revoke) → 404.
- Pin check: removing the `revokeAllowlistIfUnused` at line 180 leaves 10.0.0.99 in the per-pod set, so `validateUrl('http://10.0.0.99/v1')` would short-circuit instead of throwing — `expect(...).toThrow(SsrfError)` at test.ts:297 fails. **Branch correctly pinned.** ✓

*Test "PATCH name collision (updateProvider throws) revokes the speculative allowlist entry"* (test.ts:300–338):
- Trace: seed Throw-A (10.0.0.80) + Throw-B (10.0.0.81), then PATCH B with `{name:'Throw-A', baseUrl:'http://10.0.0.82/v1'}`.
- Route flow: `previous` = B@10.0.0.81; `addAllowedBaseUrl(10.0.0.82)`; `validateUrl` short-circuits; `updateProvider` runs `UPDATE … SET name='Throw-A', base_url=… WHERE id=B.id` → unique-constraint violation on `name` (migration 054_llm_providers.sql:9 confirms `name TEXT NOT NULL UNIQUE`) → throws → caught at route line 175 → `revokeAllowlistIfUnused('http://10.0.0.82/v1')` (DB count = 0 → revoke) → re-throws → Fastify default handler returns 500.
- Pin check: removing the `revokeAllowlistIfUnused` at line 176 leaves 10.0.0.82 allowlisted, so `expect(validateUrl('http://10.0.0.82/v1')).toThrow(SsrfError)` at test.ts:331 fails. **Branch correctly pinned.** ✓
- The two negative assertions (test.ts:333 for 10.0.0.80, test.ts:337 for 10.0.0.81) also defend against an over-broad revoke that strips other providers' origins. Good defence-in-depth.

**2. Does the doc comment accurately describe security-relevant behaviour?**

Cross-checked against `ssrf-guard.ts`:
- Claim: "validateUrl runs the protocol check BEFORE the allowlist short-circuit." Lines 238–240 (PROTOCOL_BLOCK throw) precede lines 243–245 (allowlist short-circuit return). **Accurate.** ✓
- Claim: "assertNonSsrfUrl performs DNS resolution + private-IP validation via resolveAndValidateIp." Confirmed at ssrf-guard.ts:362–364 (delegates to `validateUrlWithDns`) → ssrf-guard.ts:345–352 → `resolveAndValidateIp` at 324–335. **Accurate.** ✓
- Claim: "once a URL is allowlisted the runtime call path is exempt from DNS validation by design." `validateUrlWithDns` only calls `resolveAndValidateIp` for non-allowlisted origins (ssrf-guard.ts:349–351). **Accurate** for `validateUrlWithDns`; understates the actual situation, since the runtime LLM client doesn't call any SSRF helper at all (see Finding #1 above).
- Claim: "authenticated admin" trust model. Route preHandler at line 88 / 150 is `fastify.requireAdmin`. **Accurate.** ✓

**3. Are INFO #6 (style) and #7 (audit) genuinely fine to defer?**
- INFO #6 (`let provider`/`let updated`): both declarations exist specifically because the assignment is inside a try-block and used after; refactoring to `const` would require helper extraction with net-negative readability. **Defer is justified.** ✓
- INFO #7 (failed-create not audited): pre-existing gap that affects EVERY failure path in this file (POST validate-fail, POST create-throw, PATCH validate-fail, PATCH update-throw, PATCH 404, DELETE failures), not just this PR. Cleaner as a separate audit-coverage PR. **Defer is justified.** ✓

**4. Newly-introduced issues in test additions or comment?**
- Test 1 uses UUID `00000000-0000-4000-8000-000000000000` — valid UUIDv4 (variant `8`, version `4`), passes `IdParamsSchema`. Already used as a "definitely-doesn't-exist" UUID in `admin-user-service.test.ts:314`. Consistent. ✓
- Test 2 asserts `statusCode >= 400`. With unhandled pg unique-constraint throw, Fastify returns 500. Asserting `>= 400` is the right call given the route doesn't translate pg error 23505 to 409 (orthogonal to this PR's scope). ✓
- Test 2 comment at lines 334–336 ("the throw happens AFTER `previous` is captured but BEFORE the GC-on-success path runs") is correct; even if the GC-on-success path *did* run, it would still find Throw-B's row at 10.0.0.81 (because the UPDATE rolled back) and not revoke. Either argument supports the assertion. Minor — comment could be tighter, but isn't wrong.
- No newly-introduced issues. ✓

### Acceptance Criteria

No issue is linked to this PR (`closingIssuesReferences: []` and the body has no `Closes #NNN`). The PR title and body describe the bug + fix self-contained. Trivially-addressed work — verdict not capped on this basis.

### Story Validation

**Issue intent**: original symptom — admin POST/PATCH of a private-network LLM base URL (LM Studio on LAN, on-prem Ollama, vLLM, etc.) was rejected by the SSRF guard. Fix: allowlist before validate (mirrors Confluence flow), with proper revoke-on-failure plus baseUrl-change GC to avoid stale per-pod allowlist accretion.

**Story completion**: FULLY ADDRESSED.

**User journey**: Before — admin POSTs `http://192.168.1.50:1234/v1` → 400 SSRF blocked, can't onboard LAN LLM. After — POST allowlists then validates → 201 created; if anything downstream fails, the speculative allowlist entry is GC'd unless a concurrent sibling already owns it; PATCH likewise GCs the OLD origin once nothing references it.

**Gaps**: None.

**Evidence**:
- POST allowlist-before-validate: routes/llm/llm-providers.ts:113–115.
- POST revoke on validate-fail with concurrent-sibling guard: lines 116–125 → `revokeAllowlistIfUnused` at 43–65.
- POST revoke on create-throw with same guard: lines 127–136.
- PATCH allowlist-before-validate (same logic): lines 158–171.
- PATCH revoke on update-throw: lines 173–178.
- PATCH revoke on 404: lines 179–182.
- PATCH GC of OLD origin on successful baseUrl change: lines 186–192 with `previous` captured at line 157.
- Tests: 7 SSRF-block tests in `llm-providers.test.ts:93–339` covering happy paths, concurrent-sibling protection, PATCH GC happy/negative, plus the two new failure-branch pin tests.

### Suggested Actions

1. **[OPTIONAL]** llm-providers.ts:101-112 — consider tightening the doc comment from "exempt from DNS validation" to "exempt from runtime SSRF validation entirely" since `openai-compatible-client.ts` calls `undiciFetch` directly without invoking any SSRF helper. Defensible under the admin-trust model but a more precise statement of the threat model.

### Verdict: APPROVE (with INFO only)

**Reason**: The two new tests in `53bccef` correctly pin both the catch-on-updateProvider (line 176) and the 404-not-found (line 180) revoke branches — verified by tracing the route logic against the unique-constraint definition in migration 054. The expanded doc comment accurately describes the protocol-then-allowlist ordering in `ssrf-guard.ts:238–245` and the trust-model rationale for skipping `assertNonSsrfUrl` at create-time. INFO #6 (style `let`) and #7 (audit gap) are correctly identified as out-of-scope. Static checks pass. Story is fully addressed and traceably evidenced. No CRITICAL or WARNING findings remain — only one optional doc-wording suggestion.

---

Issue #380, opened 2026-04-28
Graph perf Phases 1 + 2 — benchmark + server-side ForceAtlas2 (follow-up to #361)


laboef1900:
Follow-up to #361 / [PR #378](https://github.com/Compendiq/compendiq-ce/pull/378). Phase 3 (per-hop neighbor cap + corpus-tiered similarity threshold) shipped in #378. The two heavier phases were intentionally deferred because they are measurement-driven and benefit from real numbers from a production-shaped corpus.

## Phase 1 — Synthetic benchmark (lands first)

Add `scripts/perf-graph-bench.ts` that:

- Generates synthetic graphs at 500 / 1000 / 2000 / 5000 nodes with realistic edge density (top-K=5 per node, score ≥ 0.6 to match the new tiered floor).
- Spins up the dev container stack and seeds Postgres directly.
- Drives a Playwright session against `/graph?focus=:id&full=1` that:
  - measures **layout convergence time** (when `cooldownTicks` hits 0)
  - samples **fps** during a programmatic pan + drag of 3 random nodes
  - records **time-to-first-paint** of the canvas and total memory (`performance.memory`)
- Outputs JSON to `perf/graph-bench-results.json`.

**Decision rule** (paste numbers in the PR comment, not just the verdict):
- If 2000 nodes hits **<30 fps drag** OR **>5 s convergence** on a mid-tier laptop → Phase 2 is justified.
- Otherwise the per-hop cap + tiered threshold from #378 are sufficient and Phase 2 stays deferred.

## Phase 2 — Server-side ForceAtlas2 (lands only if Phase 1 says so)

Take the layout cost off the request path entirely:

- New deps: `graphology`, `graphology-layout-forceatlas2`.
- Run the layout in a `worker_threads` worker OR as a BullMQ job — never inline in the Fastify event loop. `graphology-layout-forceatlas2` is synchronous and CPU-bound; for 2000 nodes a multi-second compute would block every other request in the process. Critical: **pick the off-thread option in the PR**, don't ship inline.
- New table `page_graph_layout (page_id INT PK, x REAL, y REAL, computed_at TIMESTAMPTZ)` (forward-only migration).
- Recompute trigger: end of `processDirtyPages` and `POST /api/pages/graph/refresh`. Skip eager layout for spaces with > N pages (decide N from Phase 1 numbers — likely 5000) until proven safe.
[...]

PR #378, opened 2026-04-27
feat(#361): graph perf Phase 3 — per-hop cap + tiered threshold

laboef1900:
## Summary
Phase 3 (low-risk, no new deps, no renderer decision required) of the performance work. Phases 1 (synthetic benchmark) and 2 (server-side ForceAtlas2 in `worker_threads` / BullMQ) are deferred — they need real-corpus measurements to drive the decision and benefit from shipping these caps first.

### Per-hop neighbor cap on `/graph/local`
Replaces the unbounded JOIN inside the recursive CTE with a `CROSS JOIN LATERAL ... LIMIT $3` on each BFS step, ordered by `score DESC`. Without this cap, a 3-hop expansion at a popular node in a 2000-page corpus could pull most of the corpus through the recursion. New `perHopLimit` query param (default 50, max 500) lets power users widen the search when they want targeted exploration.

### Tiered read-time similarity threshold
Producer-side threshold stays at 0.4 (data integrity — don't lose edges we may want later). At read time, `/pages/graph` applies a corpus-size-tiered floor:

| Pages | Floor |
|---|---|
| `<500` | 0.4 (small-corpus tier; no thinning) |
| `500–1999` | 0.6 (medium tier) |
| `>=2000` | 0.7 (large tier — research-recommended, prevents hairball) |

Non-similarity edges (label_overlap, explicit_link, parent_child) bypass the floor since their `score=1.0` is meaningful, not similarity-derived.

## Out of scope (next up)
- Synthetic benchmark script (`scripts/perf-graph-bench.ts`) at 500 / 1000 / 2000 / 5000 nodes — needs Playwright + FPS sampling.
- Server-side `graphology` + `graphology-layout-forceatlas2` in `worker_threads` or BullMQ so the layout cost moves off the request path; new `page_graph_layout (page_id, x, y)` table.
- Renderer decision (sigma.js v3 vs stay on react-force-graph-2d) — measurement-driven, not assumed.

Closes part of #361 · Phases 1 + 2 tracked separately

## Test plan
- [x] backend graph-route suite: 18/18 (5 new for #361)
- [x] perHopLimit forwarded into BFS as $3 binding; default 50; rejects values >500
- [x] tiered floor: 0.4 at 10-page corpus, 0.7 at 2000-page corpus
- [x] backend typecheck clean
- [ ] manual: open `/graph?focus=:id` and confirm /graph/local response shape unchanged at default `perHopLimit=50`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #361, opened 2026-04-27
Graph: 2000-node performance budget + WebGL renderer evaluation (#351d)


laboef1900:
Slice 4 of 5 from #351. Performance and scaling work. Investigation-driven — measure before swapping renderer.

## Why

The original issue assumes SVG dies around a few hundred nodes and recommends Canvas/WebGL. The page already uses Canvas via `react-force-graph-2d`. WebGL is a real upgrade for 2000+ nodes, but should be justified by a measurement, not assumed.

## Phase 1 — Benchmark current renderer

- Generate synthetic graphs at 500 / 1000 / 2000 / 5000 nodes (with realistic edge density — top-K=5 per node).
- Measure: layout convergence time, frame rate during drag, time-to-first-paint, memory.
- Decision criteria: if 2000 nodes hits <30 fps drag or >5s convergence on a mid-tier laptop, proceed to Phase 2.

## Phase 2 — Renderer swap (only if Phase 1 says so)

Candidates: `react-force-graph-3d` (WebGL, same API), `cosmograph` (GPU-accelerated), `sigma.js` (mature WebGL).
Constraints: must preserve existing features (cluster view, ego-graph, click-to-navigate, edge color/width by relationship type).

## Phase 3 — Backend caps

- Per-hop neighbor cap on `GET /api/pages/:id/graph/local`'s recursive CTE — the current 3-hop expansion can pull most of a 2000-page corpus.
- Tier the similarity threshold by corpus size (current 0.4 is permissive at scale).

## Acceptance

- [ ] Phase 1 benchmark numbers posted as a comment with the data, not just a verdict.
- [ ] If Phase 2 happens: existing tests pass against the new renderer; a11y not regressed.
- [ ] Per-hop cap and corpus-size threshold landed regardless of renderer outcome.

## Files

- `frontend/src/features/graph/GraphPage.tsx`
- `backend/src/routes/knowledge/pages-embeddings.ts` (the recursive CTE in `/graph/local`)

Blocked by: #351a, #351c (filters reduce the rendered graph and may make this moot).

Umbrella tracking issue: #351

laboef1900:
## AI Review — Issue #361

### Feasibility: Ready (with scoping clarification)

This is an investigation-driven slice with three internally-decoupled phases. Phases 1 and 3 are unambiguously implementable and low-risk. Phase 2 is correctly gated on Phase 1 evidence and should be executed as a follow-up issue rather than landed as part of this PR.

### Findings

[...]

PR #384, opened 2026-04-28
feat(#379): per-space home page picker UI

laboef1900:
Closes #379. Follow-up to #352 / #377 — the backend, migration, RBAC guard, and nav-tree filter shipped in #377; this PR wires the missing CE-side UI.

## Summary
- New `SpaceHomePicker` component dropped into Settings → Spaces rows. Trigger labels itself \"Use default home\" or \"Custom home set\" depending on whether `customHomePageId` is present.
- Built on Radix `Popover`. The popover hosts a search-style page selector scoped to the current space (uses the existing `useSearch({ q, spaceKey })` hook from `use-standalone.ts`, exactly as the issue called out), shows the resolved current home, and offers a \"Use Confluence default\" button that sends `PUT /api/spaces/:key/home` with `{ homePageId: null }` to clear the override.
- New `useSetSpaceHome` mutation hook in `use-spaces.ts` calls the new endpoint and invalidates `['spaces']` + `['pages']` on success — the backend already broadcasts a per-user-fanout cache invalidation (spaces.ts:167), so the frontend just needs to refetch and the sidebar tree picks up the new resolved `homepageId` immediately.
- RBAC gating: `usePermission('manage', 'space', spaceKey)` — the same permission the backend enforces. Non-permitted users see a disabled trigger with an explanatory tooltip; the mutation also surfaces a 403 as a `sonner` toast for the optimistic-UI fallback path the issue calls out.
- `SpacesTab.tsx` row refactor: the row was a `<button>` for the toggle, which made the picker (also a button) invalid HTML when nested. Refactored to a `<div role=\"listitem\">` with an inner `<button>` for the toggle plus the picker as a sibling that calls `stopPropagation` so its events don't bubble.

## Acceptance checklist (from #379)
- [x] Picker only appears for users who pass the same RBAC check the backend enforces — disabled trigger when `usePermission('manage','space',key)` returns `allowed:false`.
- [x] Successful save invalidates `useSpaces()` so the sidebar tree updates immediately.
- [x] Setting null restores the Confluence default; UI clearly shows the fallback (\"Currently using the Confluence default.\" copy).
- [x] Test: rendering the picker for a non-permitted user shows a disabled state.
- [x] Test: save triggers PUT with the chosen page id; clear triggers PUT with null.

## Test plan
[...]

PR #385, opened 2026-04-28
feat(#380 Phase 1): synthetic graph perf benchmark

laboef1900:
Phase 1 of #380 / #361's measurement-driven plan. Phase 2 (server-side ForceAtlas2 + page_graph_layout cache) and the sigma.js renderer migration are intentionally deferred — they only land if the numbers from this Phase 1 run say they're justified.

## Summary
- New `scripts/perf-graph-bench.ts` — generates synthetic graphs at 500/1000/2000/5000 nodes with realistic edge density (top-K=5 per node, similarity score in [0.6, 0.95] to match the corpus-tiered floor that landed in #378), drives a Playwright session against `/graph?focus=<id>&full=1`, and records the four metrics the issue asks for to `perf/graph-bench-results.json`:
  - **time-to-first-paint** — navigation → canvas mounted in DOM
  - **layout convergence** — until `requestAnimationFrame` activity stalls (force simulation has finished its `cooldownTicks=100` budget)
  - **fps during pan + drag** of 3 random nodes (2 s sample window)
  - **`performance.memory.usedJSHeapSize`** (Chrome only)
- Edge topology: ±1 / ±2 ring + a `sqrt(n)` long-range hop, so the graph has the cluster-with-bridges shape of a real Confluence corpus rather than a disconnected forest.
- Runner uses the real `/api/auth/login` flow, then injects a tiny `requestAnimationFrame` instrumentation via `addInitScript` so convergence can be detected by polling for \"no rAF in the last 350 ms\". This avoids modifying production code or the GraphPage component.
- Decision rule (issue #380, applied to the 2000-node sample): `< 30 fps drag` OR `> 5 s convergence` → Phase 2 is justified. The verdict + reasons are printed at the end of the run and stored in the JSON output.
- Pure helpers (`neighbours`, `decide`) are exported and covered by 11 vitest cases in `scripts/perf-graph-bench.test.ts`, runnable via `npx vitest run --config scripts/vitest.config.ts` — pin the edge-density invariants and the threshold logic so they regress without booting Postgres or a browser.
- Fixtures live in spaces named `PERF-BENCH-<size>` and pages with `confluence_id` prefix `perf-bench-<size>-`, so cleanup never touches real corpora.
- `perf/README.md` documents end-to-end usage (dev-stack prerequisites, CLI flags, JSON schema of the output).

## Acceptance checklist (from #380)
- [x] Phase 1 benchmark lands.
- [ ] **Numbers from a real run pasted in this PR comment** — coming once the bench has been driven against the dev stack on a mid-tier machine. (Postgres on the dev compose is on an internal-only network; see `perf/README.md` for the sidecar / port-forward options.)

## What did NOT land
- Phase 2 (server-side ForceAtlas2, `page_graph_layout` table, `worker_threads` / BullMQ off-thread compute)
- sigma.js v3 + graphology renderer migration
- Both are gated on Phase 1 numbers, per the issue.

## Test plan
- [x] `npx vitest run --config scripts/vitest.config.ts` — 11 tests pass (edge-density invariants, decision-rule thresholds, both-breach + neither-breach paths)
- [x] `npx tsc --noEmit` clean on the script + test
- [x] Seed/cleanup SQL exercised end-to-end against the running Postgres container (no schema mismatches)
- [ ] Driven against the dev stack with real numbers — to be added in a comment

### DIFF
--- a/perf/README.md
+++ b/perf/README.md
@@ -202,3 +202,108 @@ docker exec compendiq-ee-postgres-1 psql -U kb_user -d kb_creator -c \
   "DELETE FROM admin_settings WHERE setting_key='rate_limit_global_max';"
 docker restart compendiq-ee-backend-1
 ```
+
+---
+
+## Graph performance benchmark (#380 Phase 1)
+
+`scripts/perf-graph-bench.ts` measures the graph page (`/graph?focus=...&full=1`)
+under synthetic loads of 500 / 1000 / 2000 / 5000 nodes with realistic edge
+density (top-K=5 per node, similarity ≥ 0.6 to match the corpus-tiered
+floor from #378). It seeds Postgres directly, drives a Playwright session,
+and writes `perf/graph-bench-results.json`.
+
+### What it records (per size)
+
+- **time-to-first-paint** — navigation → canvas mounted in DOM
+- **layout convergence** — until `requestAnimationFrame` activity stalls
+  (force simulation has finished its `cooldownTicks=100` budget)
+- **fps during pan + drag** of 3 random nodes (2 s sample window)
+- **`performance.memory.usedJSHeapSize`** (Chrome only)
+
+### Decision rule (from the issue, applied to the 2000-node sample)
+
+- < 30 fps drag **OR** > 5 s convergence → Phase 2 (server-side ForceAtlas2)
+  is justified
+- otherwise the per-hop cap + tiered threshold from #378 are sufficient
+  and Phase 2 stays deferred
+
+The verdict is printed at the end of the run and stored in the JSON output.
+
+### Running it
+
+Postgres must be reachable from the host and the dev frontend / backend
+must be up. The fastest path on the local dev stack:
+
+```bash
+# 1. Bring up the dev stack (backend on :3052, frontend on :8081)
+docker compose -f docker/docker-compose.yml up -d --build
+
+# 2. Register a benchmark user once. The script does not touch bcrypt
+#    directly — it logs in through /api/auth/login like a real client.
+curl -X POST http://localhost:3052/api/auth/register \
+  -H 'Content-Type: application/json' \
+  -d '{"username":"benchuser","password":"benchpass123"}'
+
+# 3. Postgres on the dev compose is bound to host port 5432 by default.
+#    If yours is on the internal-only data-net (EE stack), open a one-off
+#    port-forward via `docker run --network compendiq-ee_data-net ...`
+#    or temporarily expose 5432 in the compose override.
+
+# 4. Run the bench (subset)
+npx tsx scripts/perf-graph-bench.ts --sizes=500,1000,2000
+
+# 5. Full sweep
+npx tsx scripts/perf-graph-bench.ts
+
+# 6. Drop the synthetic fixtures (the script also runs cleanup on exit
+#    unless you pass --keep-fixtures)
+npx tsx scripts/perf-graph-bench.ts --cleanup
+```
+
+### Flags / env
+
+| Flag | Default | Notes |
+|------|---------|-------|
+| `--sizes=500,1000,2000,5000` | full sweep | Comma-separated node counts |
+| `--web-url=URL` | `http://localhost:8081` | Frontend origin Playwright navigates to |
+| `--backend-url=URL` | `http://localhost:3051` | Backend used for `/api/auth/login` |
+| `--username=NAME` | `benchuser` | Must already exist (register once) |
+| `--password=PASS` | `benchpass123` | |
+| `--headless=true|false` | `true` | `false` to watch the browser drive |
+| `--keep-fixtures` | off | Skip cleanup at the end of the run |
+| `--cleanup` | — | Drop fixtures and exit (no measurement) |
+| `POSTGRES_URL` env | `postgresql://kb_user:changeme-postgres@localhost:5432/kb_creator` | |
+
+### Unit tests for the helpers
+
+The pure helpers (`neighbours`, `decide`) carry their own test suite:
+
+```bash
+npx vitest run --config scripts/vitest.config.ts
+```
+
+So the threshold logic and the edge-density invariants can be regressed
+without booting Postgres or a browser.
+
+### Result format
+
+```json
+{
+  "schemaVersion": 1,
+  "issue": "#380",
+  "runAt": "2026-04-28T13:00:00.000Z",
+  "config": { "sizes": [...], "edgesPerNode": 5, "minScore": 0.6, "maxScore": 0.95 },
+  "samples": [
+    { "size": 500, "pageCount": 500, "edgeCount": 1234,
+      "timeToFirstPaintMs": 380, "layoutConvergenceMs": 720,
+      "fpsDuringInteraction": 60, "jsHeapUsedMb": 92.4 },
+    ...
+  ],
+  "verdict": { "phase2Justified": false, "reasons": [] }
+}
+```
+
+Paste the four samples + verdict into the PR comment — that's the
+artefact the issue asks for ("paste numbers in the PR comment, not just
+the verdict").
--- a/perf/README.md
+++ b/perf/README.md
@@ -274,6 +274,7 @@ npx tsx scripts/perf-graph-bench.ts --cleanup
 | `--keep-fixtures` | off | Skip cleanup at the end of the run |
 | `--cleanup` | — | Drop fixtures and exit (no measurement) |
 | `POSTGRES_URL` env | `postgresql://kb_user:changeme-postgres@localhost:5432/kb_creator` | |
+| `REDIS_URL` env | unset | Optional. When set, the script flushes the bench user's RBAC space-list cache after seeding so consecutive runs with different `--sizes` don't read a stale 60 s-cached value. |
 
 ### Unit tests for the helpers
 
@@ -285,6 +286,25 @@ npx vitest run --config scripts/vitest.config.ts
 
 So the threshold logic and the edge-density invariants can be regressed
 without booting Postgres or a browser.
+
+### Caveats / known gotchas
+
+- **RBAC cache TTL**: `getUserAccessibleSpaces` caches results in Redis for
+  60 s (per `rbac-service.ts`). If the bench user has been seeded with a
+  *different* set of spaces in the previous 60 s (e.g. you ran the bench
+  with `--sizes=500,1000` and immediately re-ran with `--sizes=2000,5000`),
+  the cached space list may be stale and the global graph endpoint will
+  return rows from spaces that aren't in this run's seed. Either wait
+  60 s between runs that mutate the bench user's space access, or flush
+  the Redis key (`compendiq:cache:<userId>:spaces:list`) between runs.
+- **Reduced motion**: the script pins `reducedMotion: 'no-preference'`
+  on the Playwright context. Without that, the GraphPage's
+  `cooldownTicks={prefersReducedMotion ? 0 : 100}` short-circuits and
+  the convergence sample bottoms out at 0 ms regardless of size.
+- **Postgres reachability**: dev compose binds Postgres on the
+  `data-net` (internal-only on the EE stack, exposed on the CE stack).
+  If your Postgres is internal-only, run the bench inside a sidecar
+  container that joins the data network, or temporarily expose 5432.
 
 ### Result format
 
--- a/perf/README.md
+++ b/perf/README.md
@@ -267,7 +267,7 @@ npx tsx scripts/perf-graph-bench.ts --cleanup
 |------|---------|-------|
 | `--sizes=500,1000,2000,5000` | full sweep | Comma-separated node counts |
 | `--web-url=URL` | `http://localhost:8081` | Frontend origin Playwright navigates to |
-| `--backend-url=URL` | `http://localhost:3051` | Backend used for `/api/auth/login` |
+| `--backend-url=URL` | `http://localhost:3052` | Backend used for `/api/auth/login` (matches dev compose's `BACKEND_HOST_PORT` default) |
 | `--username=NAME` | `benchuser` | Must already exist (register once) |
 | `--password=PASS` | `benchpass123` | |
 | `--headless=true|false` | `true` | `false` to watch the browser drive |
--- a/scripts/perf-graph-bench.test.ts
+++ b/scripts/perf-graph-bench.test.ts
@@ -1,0 +1,103 @@
+import { describe, it, expect } from 'vitest';
+import { neighbours, decide, EDGES_PER_NODE, type BenchSample } from './perf-graph-bench';
+
+describe('perf-graph-bench helpers (#380 Phase 1)', () => {
+  describe('neighbours', () => {
+    it('returns at most EDGES_PER_NODE neighbours for any node', () => {
+      for (const n of [10, 100, 500, 2000]) {
+        for (const i of [0, Math.floor(n / 2), n - 1]) {
+          expect(neighbours(i, n).length).toBeLessThanOrEqual(EDGES_PER_NODE);
+        }
+      }
+    });
+
+    it('never returns the source node as its own neighbour', () => {
+      for (const n of [10, 100, 1000]) {
+        for (let i = 0; i < n; i++) {
+          expect(neighbours(i, n)).not.toContain(i);
+        }
+      }
+    });
+
+    it('produces unique neighbours per call', () => {
+      const result = neighbours(50, 1000);
+      expect(new Set(result).size).toBe(result.length);
+    });
+
+    it('wraps around so first/last nodes still get neighbours', () => {
+      // Node 0 should still produce neighbours via the modular wrap
+      // (offset -1 wraps to n-1) — otherwise the head of the graph would
+      // be a sink.
+      expect(neighbours(0, 100).length).toBeGreaterThan(0);
+      expect(neighbours(99, 100).length).toBeGreaterThan(0);
+    });
+
+    it('includes a long-range hop (~sqrt(n)) so the graph is not just a ring', () => {
+      // For n=2000, sqrt is ~44 — neighbours(0, 2000) should include
+      // either 44 or, where collisions happen, a sibling. We just assert
+      // that one of the neighbours is far away from i.
+      const out = neighbours(0, 2000);
+      const hasLongRange = out.some((j) => Math.min(j, 2000 - j) >= 10);
+      expect(hasLongRange).toBe(true);
+    });
+  });
+
+  describe('decide', () => {
+    function sample(overrides: Partial<BenchSample>): BenchSample {
+      return {
+        size: 2000,
+        pageCount: 2000,
+        edgeCount: 5000,
+        centerPageId: 1,
+        timeToFirstPaintMs: 500,
+        layoutConvergenceMs: 2000,
+        fpsDuringInteraction: 60,
+        jsHeapUsedMb: 100,
+        ...overrides,
+      };
+    }
+
+    it('says Phase 2 is NOT justified when 2000-node sample is healthy', () => {
+      const verdict = decide([sample({ size: 2000 })]);
+      expect(verdict.phase2Justified).toBe(false);
+      expect(verdict.reasons).toEqual([]);
+    });
+
+    it('flags fps < 30 at 2000 nodes', () => {
+      const verdict = decide([sample({ size: 2000, fpsDuringInteraction: 22 })]);
+      expect(verdict.phase2Justified).toBe(true);
+      expect(verdict.reasons.some((r) => r.includes('fps'))).toBe(true);
+    });
+
+    it('flags convergence > 5000 ms at 2000 nodes', () => {
+      const verdict = decide([sample({ size: 2000, layoutConvergenceMs: 6500 })]);
+      expect(verdict.phase2Justified).toBe(true);
+      expect(verdict.reasons.some((r) => r.includes('convergence'))).toBe(true);
+    });
+
+    it('reports both reasons when both thresholds breach', () => {
+      const verdict = decide([
+        sample({ size: 2000, layoutConvergenceMs: 7000, fpsDuringInteraction: 18 }),
+      ]);
+      expect(verdict.phase2Justified).toBe(true);
+      expect(verdict.reasons.length).toBe(2);
+    });
+
+    it('returns phase2Justified=false with an explanation when no 2000-node sample exists', () => {
+      const verdict = decide([sample({ size: 1000 })]);
+      expect(verdict.phase2Justified).toBe(false);
+      expect(verdict.reasons[0]).toMatch(/2000-node/);
+    });
+
+    it('uses the 2000-node sample even when other sizes also breach', () => {
+      // 5000 with bad fps must NOT trigger Phase 2 on its own — the
+      // decision rule is anchored on the 2000 sample (issue #380).
+      const verdict = decide([
+        sample({ size: 1000, fpsDuringInteraction: 55 }),
+        sample({ size: 2000, fpsDuringInteraction: 55, layoutConvergenceMs: 2500 }),
+        sample({ size: 5000, fpsDuringInteraction: 12 }),
+      ]);
+      expect(verdict.phase2Justified).toBe(false);
+    });
+  });
+});
--- a/scripts/perf-graph-bench.test.ts
+++ b/scripts/perf-graph-bench.test.ts
@@ -48,7 +48,7 @@ describe('perf-graph-bench helpers (#380 Phase 1)', () => {
         size: 2000,
         pageCount: 2000,
         edgeCount: 5000,
-        centerPageId: 1,
+        samplePageId: 1,
         timeToFirstPaintMs: 500,
         layoutConvergenceMs: 2000,
         fpsDuringInteraction: 60,
--- a/scripts/perf-graph-bench.test.ts
+++ b/scripts/perf-graph-bench.test.ts
@@ -1,5 +1,11 @@
 import { describe, it, expect } from 'vitest';
-import { neighbours, decide, EDGES_PER_NODE, type BenchSample } from './perf-graph-bench';
+import {
+  neighbours,
+  decide,
+  benchInitScriptBody,
+  EDGES_PER_NODE,
+  type BenchSample,
+} from './perf-graph-bench';
 
 describe('perf-graph-bench helpers (#380 Phase 1)', () => {
   describe('neighbours', () => {
@@ -98,6 +104,95 @@ describe('perf-graph-bench helpers (#380 Phase 1)', () => {
         sample({ size: 5000, fpsDuringInteraction: 12 }),
       ]);
       expect(verdict.phase2Justified).toBe(false);
+    });
+  });
+
+  // ---- benchInitScriptBody single-shot guarantee ----
+  // Playwright's addInitScript replays the script on every navigation,
+  // and the prior bug shipped a fresh rAF wrapper layer on each replay.
+  // The fix relies on a window-level `__benchInstalled` flag. These
+  // tests pin that contract in a fake-DOM so a regression can't slip
+  // through without the full Playwright lane catching it.
+  describe('benchInitScriptBody single-shot guarantee', () => {
+    function buildFakeWindow(): {
+      rafCalls: Array<(t: number) => void>;
+      windowProxy: Record<string, unknown>;
+    } {
+      const rafCalls: Array<(t: number) => void> = [];
+      const baseRaf = (cb: (t: number) => void) => {
+        rafCalls.push(cb);
+        return rafCalls.length;
+      };
+      const ls = new Map<string, string>();
+      const windowProxy: Record<string, unknown> = {
+        requestAnimationFrame: baseRaf,
+        localStorage: {
+          setItem: (k: string, v: string) => ls.set(k, v),
+          getItem: (k: string) => ls.get(k) ?? null,
+        },
+        performance: { now: () => 0 },
+      };
+      return { rafCalls, windowProxy };
+    }
+
+    function runScript(windowProxy: Record<string, unknown>) {
+      // Simulate Playwright's evaluation: the script body sees the page
+      // window as `window`. Bind a tiny shim so the function body's
+      // `window` references resolve to our fake.
+      const fn = new Function(
+        'window',
+        'localStorage',
+        'performance',
+        `(${benchInitScriptBody.toString()})({ token: 'tok' });`,
+      );
+      fn(windowProxy, windowProxy.localStorage, windowProxy.performance);
+    }
+
+    it('wraps requestAnimationFrame exactly once across multiple replays', () => {
+      const { windowProxy } = buildFakeWindow();
+      const original = windowProxy.requestAnimationFrame;
+
+      // First replay (initial page load).
+      runScript(windowProxy);
+      const afterFirst = windowProxy.requestAnimationFrame;
+      expect(afterFirst).not.toBe(original);
+      expect(windowProxy.__benchInstalled).toBe(true);
+
+      // Second replay (a goto in the same page lifetime).
+      runScript(windowProxy);
+      expect(windowProxy.requestAnimationFrame).toBe(afterFirst);
+
+      // Third replay (next iteration of the size sweep).
+      runScript(windowProxy);
+      expect(windowProxy.requestAnimationFrame).toBe(afterFirst);
+    });
+
+    it('resets the rAF counter on every replay', () => {
+      const { windowProxy } = buildFakeWindow();
+
+      runScript(windowProxy);
+      const bench1 = windowProxy.__bench as Record<string, number>;
+      bench1.rafCount = 42;
+      bench1.lastRaf = 9999;
+
+      runScript(windowProxy);
+      const bench2 = windowProxy.__bench as Record<string, number>;
+      expect(bench2.rafCount).toBe(0);
+      expect(bench2.lastRaf).toBe(0);
+      expect(bench2.startedAt).toBe(0);
+    });
+
+    it('seeds the auth token into localStorage so the SPA boots authenticated', () => {
+      const { windowProxy } = buildFakeWindow();
+      runScript(windowProxy);
+
+      const stored = (windowProxy.localStorage as { getItem: (k: string) => string | null }).getItem(
+        'compendiq-auth',
+      );
+      expect(stored).toBeTruthy();
+      const parsed = JSON.parse(stored!);
+      expect(parsed.state.accessToken).toBe('tok');
+      expect(parsed.state.isAuthenticated).toBe(true);
     });
   });
 });
--- a/scripts/perf-graph-bench.ts
+++ b/scripts/perf-graph-bench.ts
@@ -1,0 +1,621 @@
+#!/usr/bin/env npx tsx
+/**
+ * Graph performance benchmark — issue #380 Phase 1.
+ *
+ * Generates synthetic graphs at the four target sizes (default
+ * 500/1000/2000/5000) with realistic edge density (top-K=5 per node, score
+ * >= 0.6 to match the corpus-tiered floor that landed in #378), drives a
+ * Playwright session against `/graph?focus=<id>&full=1`, and records:
+ *
+ *   - layout convergence time (rAF activity stalls)
+ *   - fps during a programmatic pan + drag of 3 random nodes
+ *   - time-to-first-paint (canvas mounted in DOM)
+ *   - JS heap usage from `performance.memory`
+ *
+ * Output: `perf/graph-bench-results.json`. The script applies the issue's
+ * decision rule and prints whether Phase 2 is justified.
+ *
+ * Usage:
+ *   docker compose -f docker/docker-compose.yml up -d --build   # bring up the dev stack
+ *   npx tsx scripts/perf-graph-bench.ts                         # full sweep
+ *   npx tsx scripts/perf-graph-bench.ts --sizes=500,1000        # subset
+ *   npx tsx scripts/perf-graph-bench.ts --cleanup               # remove fixtures
+ *
+ * Env / flags:
+ *   POSTGRES_URL          (default postgresql://kb_user:changeme-postgres@localhost:5432/kb_creator)
+ *   --web-url=URL         (default http://localhost:8081)
+ *   --backend-url=URL     (default http://localhost:3051)
+ *   --username=...        (default benchuser)
+ *   --password=...        (default benchpass123)
+ *   --headless=true|false (default true; set false to watch the browser)
+ *   --keep-fixtures       (skip cleanup at the end so re-runs are faster)
+ *
+ * The synthetic fixtures live in spaces named `PERF-BENCH-<size>` and pages
+ * with `confluence_id` prefix `perf-bench-<size>-`, so cleanup never
+ * touches real corpora. The benchmark user is granted access via the
+ * existing RBAC seed (`user_space_selections` + `space_admin` role), so
+ * the same fixtures show up in the `/api/pages/graph` endpoint without
+ * any RBAC monkey-patching.
+ */
+
+import pg from 'pg';
+import { chromium } from 'playwright';
+import { mkdir, writeFile } from 'node:fs/promises';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const { Pool } = pg;
+
+// ── Configuration ───────────────────────────────────────────────────────────
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const REPO_ROOT = resolve(__dirname, '..');
+const RESULTS_PATH = resolve(REPO_ROOT, 'perf/graph-bench-results.json');
+
+const POSTGRES_URL =
+  process.env.POSTGRES_URL ??
+  'postgresql://kb_user:changeme-postgres@localhost:5432/kb_creator';
+
+interface Args {
+  sizes: number[];
+  webUrl: string;
+  backendUrl: string;
+  username: string;
+  password: string;
+  cleanup: boolean;
+  keepFixtures: boolean;
+  headless: boolean;
+}
+
+function parseArgs(): Args {
+  const argv = process.argv.slice(2);
+  const get = (name: string, fallback: string) => {
+    const arg = argv.find((a) => a.startsWith(`--${name}=`));
+    return arg ? arg.slice(name.length + 3) : fallback;
+  };
+  const flag = (name: string) => argv.includes(`--${name}`);
+  const sizes = get('sizes', '500,1000,2000,5000')
+    .split(',')
+    .map((s) => parseInt(s.trim(), 10))
+    .filter((n) => Number.isFinite(n) && n > 0);
+  return {
+    sizes,
+    webUrl: get('web-url', 'http://localhost:8081').replace(/\/$/, ''),
+    backendUrl: get('backend-url', 'http://localhost:3051').replace(/\/$/, ''),
+    username: get('username', 'benchuser'),
+    password: get('password', 'benchpass123'),
+    cleanup: flag('cleanup'),
+    keepFixtures: flag('keep-fixtures'),
+    headless: get('headless', 'true') !== 'false',
+  };
+}
+
+// ── Edge density: top-K=5 per node, score floor 0.6 ─────────────────────────
+
+export const EDGES_PER_NODE = 5;
+export const MIN_SCORE = 0.6;
+export const MAX_SCORE = 0.95;
+
+function randomScore(): number {
+  return MIN_SCORE + Math.random() * (MAX_SCORE - MIN_SCORE);
+}
+
+/**
+ * Deterministic neighbour generator — exported for unit tests so the
+ * synthetic graph density can be verified without touching Postgres.
+ *
+ * A node connects to its ±1, ±2 ring plus a `sqrt(n)` long-range hop.
+ * That mimics the cluster-with-bridges shape of a real Confluence corpus
+ * and keeps the centre node reachable from most others, which is what a
+ * focused (`?focus=`) view actually exercises.
+ */
+export function neighbours(i: number, n: number): number[] {
+  const offsets = [1, -1, 2, -2, Math.floor(Math.sqrt(n))];
+  const seen = new Set<number>();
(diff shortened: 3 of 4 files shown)


laboef1900 (commented):
## Automated PR Review

**PR**: #385 -- feat(#380 Phase 1): synthetic graph perf benchmark
**Branch**: feature/380-graph-perf-bench -> dev
**Diff size**: 842 lines (+842/-0) across 4 files
**Reviewed at**: 2026-04-28 15:50 UTC

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript (workspaces) | PASS | contracts + backend + frontend clean |
| TypeScript (script direct) | PASS | `scripts/perf-graph-bench.ts` + test file clean under strict |
| Lint | PASS | backend + frontend, no warnings |
| Backend Tests | INCONCLUSIVE | Suite ran 96+ files passing then timed out at 600s; competing for shared test PG with another concurrent agent. No PR-attributable failures observed. |
| Frontend Tests | PASS | 2004/2004 passed across 167 test files |
| Production Build | PASS | Vite + tsc, all chunks built |
| Script Unit Tests | PASS | 11/11 (`npx vitest run --config scripts/vitest.config.ts`) |
| npm audit | PASS | 0 high, 0 critical (3 moderate, all pre-existing) |
| Branch Target | PASS | `feature/* -> dev` |

### Docker App Test

| Check | Status |
|-------|--------|
| All checks | SKIPPED |

Docker tests skipped: ports 5433 (Postgres) held by the concurrent dev/test stack. Consistent with the prior reviews on #383 and #384.

### E2E Browser Tests

| Project | Status |
|---------|--------|
| All projects | SKIPPED |

E2E tests skipped: Docker phase skipped, so no fresh stack to drive. The Phase 1 bench itself has not been executed; that's the deferred AC the PR description acknowledges.

### Code Review Findings

| 1 | CRITICAL | scripts/perf-graph-bench.ts:343-374 | `addInitScript` is registered inside `runOne`, which is called once per size. Playwright accumulates init scripts and replays ALL of them on every subsequent navigation, in registration order. Each replay re-wraps `window.requestAnimationFrame`, so iteration N has N nested wrappers and `data.rafCount` increments N times per real frame. For the 4-size sweep this means iter 1 reports 1× real fps, iter 2 reports 2×, iter 3 (the 2000-node decision sample) reports 3×, iter 4 reports 4×. The threshold `fpsDuringInteraction < 30` therefore triggers only when real fps is < 10 at 2000 nodes — so a real 20 fps reads as 60 fps and Phase 2 will be wrongly declared "not justified". This breaks the decision rule the entire script is built around. Fix: register the init script once outside the loop, OR call `context.removeInitScript(...)` (Playwright 1.44+) / open a fresh page per size, OR cache `window.__bench_originalRaf` on first wrap and skip rewrapping. |
| 2 | CRITICAL | scripts/perf-graph-bench.ts:376 | Bench URL is `/graph?focus=<id>&full=1`. On the PR's current base, `full=1` is unrecognised by `GraphPage.tsx` — the only params read are `focus`, `view`, `space`. Even on `origin/dev` (where #375 lands `?full=1`) the routing is `const wantsGlobal = !focusPageId && showFullGraph; const activeQuery = focusPageId ? localQuery : (wantsGlobal ? globalQuery : null);` — i.e. when `focus` is set, `full=1` is **ignored** and the page calls `useLocalGraphData(focusPageId)` which hits `/pages/{id}/graph/local?hops=2`. With top-K=5 neighbours the 2-hop ego graph is ~30 nodes regardless of the seeded corpus size. **The benchmark does not measure rendering at 500/1000/2000/5000 nodes — it measures a ~30-node ego graph four times.** The decision rule is anchored on the 2000-node sample, so the verdict is meaningless. Fix: drop `focus=` from the URL when you want the full graph (`/graph?full=1` after rebasing on dev), and adjust the seeded fixture topology so a single space holds the target node count instead of four cumulative spaces. |
| 3 | CRITICAL | scripts/perf-graph-bench.ts:170-181 | `grantSpaceAccess` writes to `user_space_selections`, but `backend/src/core/services/rbac-service.ts:getUserAccessibleSpaces` explicitly notes "This does NOT query user_space_selections. ... RBAC space access is determined solely by space_role_assignments." On a dev stack where `benchuser` is not the first registered user (i.e. not auto-promoted to admin), `getUserAccessibleSpaces(benchuser)` returns `[]`, so `/api/pages/graph` and `/pages/{id}/graph/local` short-circuit to `{ nodes: [], edges: [] }` and the bench renders an empty graph. The script's docstring (lines 35-38) claims the user is granted access via `user_space_selections + space_admin role`, but neither table grants access in current code. Fix: insert into `space_role_assignments(space_key, principal_type='user', principal_id=userId, role_id=<some_role>)` for each PERF-BENCH-* space, OR promote `benchuser` to admin role (matches "first user is admin" but unsafe to do silently on shared stacks). |
| 4 | WARNING | scripts/perf-graph-bench.ts:85, perf/README.md:268-272 | Script default `--backend-url=http://localhost:3051` doesn't match `docker/docker-compose.yml`, which maps `BACKEND_HOST_PORT:-3052` -> 3051 in-container. The README correctly uses `:3052` for the manual register step, but a plain `npx tsx scripts/perf-graph-bench.ts` against the dev compose will fail login with ECONNREFUSED on 3051. Either raise the script default to 3052 or document the override prominently. The Postgres default `localhost:5432` likewise doesn't have a host mapping in `docker/docker-compose.yml` (postgres is on internal `data-net`); the README does mention the sidecar option lower down, but the script default suggests it'll just work locally. |
| 5 | WARNING | scripts/perf-graph-bench.ts (whole file) | The PR base (commit ecec1132) is several commits behind `origin/dev`. It is missing PR #378 (the per-hop cap whose effect Phase 1 is supposed to validate), PR #371/#372/#375/#376 (graph empty states, explicit_link edges, ego-graph default with `?full=1` URL handling, parent_child edges). Running the bench against this base would measure a graph route that lacks the very Phase-3 fix it's supposed to bracket. **The PR must be rebased onto current dev before the benchmark numbers carry any signal.** |
| 6 | WARNING | scripts/perf-graph-bench.ts:438-450 | Comment claims this section drags 3 "random nodes", but the implementation just issues mouse-down/move/up at random offsets around the canvas centre. It almost never lands on an actual node (canvas hit-detection requires the `react-force-graph-2d` node renderer to register the click), so what's measured is an empty-canvas drag, not a node drag. Pan-on-empty has very different rendering cost than node-drag (which forces simulation re-tick on every mousemove). The fps metric the issue calls out — "fps during pan + drag of 3 random nodes" — is half-measured. Fix: read node screen positions from `graphRef.current.graph2ScreenCoords(node.x, node.y)` via `page.evaluate`, OR expose the first 3 node coordinates via a `data-bench-positions` attribute the bench can read, OR accept that this metric measures pan only and drop "drag of 3 random nodes" from the docstring. |
| 7 | WARNING | scripts/perf-graph-bench.ts:392-407 | `layoutConvergenceMs` returns `Math.max(0, data.lastRaf - startedAt)` but on a quiet (no-rAF-since-start) page `data.lastRaf` could be 0 and `startedAt` could be set in a later browser tick, producing a negative elapsed that gets clamped to 0 -- which then reads as "instant convergence" and silently bypasses the >5000ms threshold. The fallback at line 405 (`return performance.now() - startedAt`) only fires after the full 30s timeout, but on small graphs that complete during DOMContentLoaded the early-success path can return 0. Add a sanity floor (e.g., return `Math.max(50, ...)`) or use `navStart` instead of `startedAt` for the lower bound. |
| 8 | WARNING | scripts/perf-graph-bench.ts:130-289 | The whole `seedSize` function is in a single transaction with `ON CONFLICT DO NOTHING` on edge inserts but `seenPair` deduplication on the JS side. If the same size is seeded twice (idempotency), the second run starts with `cleanupSize` -> all rows deleted -> fresh page IDs (SERIAL keeps incrementing). That works, but means re-running with `--keep-fixtures` between runs and then without it again will leave orphaned `page_relationships` rows whose `page_id_1`/`page_id_2` no longer match deleted pages -- except CASCADE handles that. Ok in practice, but worth a comment that `--keep-fixtures` is for inner re-runs only. |
| 9 | INFO | scripts/perf-graph-bench.ts:77 | `flag(name)` does exact-match on `argv.includes(--name)`, so `--keep-fixtures=true` would silently parse as false (no error). Minor UX nit. |
| 10 | INFO | scripts/perf-graph-bench.ts:215 | `cid = perf-bench-${size}-${i.padStart(6,'0')}` produces 7-digit strings for size=5000 (i=4999), so the prefix is fine, but the comment at line 33-38 says prefix is `perf-bench-<size>-` not `perf-bench-<size>-<6-digit-i>` -- minor doc drift. |
| 11 | INFO | scripts/perf-graph-bench.ts:478 | `Math.round(fpsDuringInteraction * 10) / 10` rounds to 1 decimal place (e.g., 22.7), but the threshold check is `< 30`. Value of 29.95 rounds to 30.0 and the condition fails. Use raw value for the comparison and round only for the printed line. |
| 12 | INFO | scripts/perf-graph-bench.ts:621 | Auto-run guard uses `import.meta.url === file://${process.argv[1]}` OR `process.argv[1]?.endsWith('perf-graph-bench.ts')`. The endsWith branch will also fire when running the test file via `vitest` if vitest happens to expose the source as argv[1] in some configurations. Vitest doesn't currently do this, but the path-suffix check is fragile. The `import.meta.url === ...` check is sufficient on its own. |

### Acceptance Criteria (Issue #380)

| AC-1 | Phase 1 benchmark lands with numbers in the PR comment, not just a verdict | PARTIAL | Script lands with full structure, schema, decision rule, and 11 unit tests. Numbers from a real run are explicitly NOT in the PR yet (PR description acknowledges this). The script in its current form would produce numbers that do NOT measure the size dimension (see findings 1, 2, 3) — running it as-is would produce inflated, ego-graph-only fps numbers. |
| AC-2 (Phase 2) | If Phase 2 happens: layout job runs off the main thread; `page_graph_layout` populates; client reports zero simulation cost on next load | N/A | Intentionally deferred — correct per the issue's measurement-driven plan. |
| AC-3 (renderer) | If renderer swap happens: existing tests pass, behaviours preserved, bundle delta documented | N/A | Intentionally deferred. |

### Story Validation (Issue #380)

**Issue intent**: Land a synthetic-corpus benchmark that measures the four metrics (TTFP, layout convergence, drag fps, heap) at 500/1000/2000/5000 nodes against the running graph page, and apply the 2000-node decision rule (< 30 fps drag OR > 5 s convergence -> Phase 2 justified). Numbers, not just verdict, in the PR comment.

**Story completion**: PARTIALLY ADDRESSED

**User journey**: Before: no data on whether Phase 2 (server-side ForceAtlas2) is justified — the Phase 3 caps from #378 may be enough on their own. After: a script exists that *claims* to measure the four metrics, but in practice (a) doesn't exercise the size dimension because the focus-URL routes through the 2-hop ego endpoint, (b) inflates fps by N at iter N due to nested rAF wrappers, (c) returns empty graphs on non-admin users because RBAC isn't actually granted, and (d) targets a stale base that doesn't yet contain #378. So the benchmark in its current form would not deliver the data the issue is asking for.

**Gaps**:
- The bench URL (`/graph?focus=<id>&full=1`) doesn't trigger the full graph — `?full=1` is ignored when `focus` is set, even after rebase on dev.
- RBAC seed targets the wrong table (`user_space_selections` instead of `space_role_assignments`).
- rAF wrapper accumulation across iterations breaks the decision rule's anchor metric.
- PR is on a stale base, missing the #378 per-hop cap that Phase 1 is meant to bracket.
- "drag of 3 random nodes" is actually drag-on-empty-canvas (no node coordinates resolved).

**Evidence**:
- `scripts/perf-graph-bench.ts:376` constructs the broken URL.
- `scripts/perf-graph-bench.ts:343-374` registers a fresh init script per iteration.
- `scripts/perf-graph-bench.ts:170-181` writes to the wrong RBAC table (verified against `backend/src/core/services/rbac-service.ts`).
- `git log ecec1132..origin/dev` shows 6 commits missing including #378.

The structure (vitest unit tests for pure helpers, JSON output schema, `decide()` separated from `runOne`, README usage, fixture isolation by space-key prefix) is good craftsmanship and worth keeping. The bench just needs the four issues above fixed before it can produce a usable verdict.

### Suggested Actions

1. **[REQUIRED]** Fix the rAF wrapper accumulation so per-iteration fps is not multiplied by iteration count. Easiest: register `addInitScript` once outside the size loop using a fresh wrapper that checks for an existing `__bench_originalRaf` global before re-wrapping. (Finding 1)
2. **[REQUIRED]** Drop `&full=1` from the URL when `focus=` is set, or remove `focus=` so `full=1` actually takes effect after rebasing on dev. Reseed each size into its OWN space and grant access to only that space per iteration so the global view returns exactly N nodes. (Finding 2)
3. **[REQUIRED]** Replace `INSERT INTO user_space_selections` with `INSERT INTO space_role_assignments(space_key, principal_type='user', principal_id, role_id=<viewer-or-equivalent>)`. Look up the role by name first (e.g. `SELECT id FROM roles WHERE name='space_viewer'`). (Finding 3)
4. **[REQUIRED]** Rebase the branch onto current `origin/dev` so the bench measures the per-hop cap from #378, not the unbounded recursion that pre-dates it. (Finding 5)
5. **[SHOULD FIX]** Update the script default `--backend-url` to `http://localhost:3052` to match the dev compose, or add a guard that probes both ports and errors with a clear message. (Finding 4)
6. **[SHOULD FIX]** Resolve real node screen coordinates (via `graphRef.graph2ScreenCoords` exposed through `page.evaluate`) before the drag loop, or change the docstring to match what's actually measured. (Finding 6)
7. **[SHOULD FIX]** Guard `layoutConvergenceMs` against the early-success / negative-elapsed corner case. (Finding 7)
8. **[CONSIDER]** Address the minor INFO findings (9-12) when touching this code. They're each isolated 1-2 line fixes.

### Verdict: REQUEST_CHANGES

**Reason**: Three independently CRITICAL issues mean the script as written does not measure what the issue asks for: rAF wrappers compound across iterations and inflate fps by Nx (finding 1); the URL routes to the 2-hop ego endpoint regardless of seeded corpus size (finding 2); RBAC isn't actually granted on non-admin users (finding 3). A run today would either error out (empty graph for non-admins) or produce numbers that flatter the existing system (false-negative on Phase 2 justification). The framework is good — unit-tested helpers, decision rule isolated, JSON schema versioned, fixture cleanup by prefix — but it needs to actually exercise the size dimension before any verdict it produces can drive a Phase-2 decision.

---

laboef1900:
Addressed the review's three CRITICAL findings + the three WARNINGs in `40bf559`. Branch is rebased onto current `origin/dev` (so #378's per-hop cap is in scope).

**Critical fixes**

1. **rAF wrapper accumulation** — `addInitScript` is now hoisted into a dedicated `installBenchInit()` called once outside the iteration loop, with an inner `__benchInstalled` guard against double-wrapping. The per-iteration counter reset moved to a tiny `page.evaluate` so each sample starts clean. The 2000-node decision sample is no longer read at 3× the real frame rate.
2. **`focus=…&full=1` silently downgraded to the ego graph** — bench now navigates to `/graph?full=1` (no `focus=`), so on origin/dev's `wantsGlobal = !focusPageId && showFullGraph` route it actually hits `useGraphData` and renders the full seeded corpus. The seeded centre id is retained as `samplePageId` in the JSON output for traceability but doesn't drive routing.
3. **Wrong RBAC table** — `grantSpaceAccess` now inserts an `editor` role into `space_role_assignments` (the table `getUserAccessibleSpaces` actually consults; rbac-service.ts states "This does NOT query user_space_selections"). `cleanupSize` removes those assignments so re-runs don't accumulate stale rows. Verified the SQL end-to-end against the running Postgres container.

**Warning fixes**

- Rebased onto origin/dev (PR #378 + #375 etc. now in the merge base).
- Default `--backend-url` is `:3052` to match the dev compose's `BACKEND_HOST_PORT`. Table in `perf/README.md` updated.
- Comments around the pan + drag gestures reframed: the metric is "rAF rate during interaction", not "node drag mechanics". The renderer's internal node positions aren't exposed to the page, so the gestures intentionally don't target specific nodes — the fps reading is dominated by simulation cost regardless.

**Verification**

- `npx vitest run --config scripts/vitest.config.ts` — 11/11 pass
- `npx tsc --noEmit` clean on the script + test
- `npx tsx scripts/perf-graph-bench.ts` parses cleanly (only fails on Postgres ECONNREFUSED, expected on internal-only network)
- Seed/RBAC SQL exercised against running Postgres via `docker exec`: `space_role_assignments` insert + role lookup return the expected role_id, full transaction round-trips clean.

The bench has not been driven against a fully isolated dev stack yet — that's the remaining open item; will paste the four-size numbers + verdict in a subsequent comment once a clean run completes.

laboef1900 (commented):
## Automated PR Review (Re-Review)

**PR**: #385 -- feat(#380 Phase 1): synthetic graph perf benchmark
**Branch**: feature/380-graph-perf-bench -> dev
**Diff size**: ~900 lines (+900/-0) across 4 files
**Re-reviewing commit**: `40bf559 fix(#380): address PR #385 review findings`
**Reviewed at**: 2026-04-28

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | `npm run typecheck` clean (contracts + backend + frontend) |
| Lint | PASS | `npm run lint` clean (only pre-existing boundaries v5->v6 warning, unrelated) |
| Unit tests (perf-graph-bench helpers) | PASS | 11/11 passed via `npx vitest run --config scripts/vitest.config.ts` |
| Backend / Frontend full test suite | SKIPPED | Suite needs running Postgres; bench changes are confined to `scripts/**` and `perf/README.md` and don't touch backend/frontend code |
| Production Build | NOT RUN | Bench is a standalone tsx script; build pipeline doesn't include it |
| npm audit | INFO | 4 moderate vulnerabilities, no high/critical |
| Branch Target | PASS | `feature/380-graph-perf-bench -> dev` (matches CLAUDE.md policy) |

### Docker App Test

| Check | Status |
|-------|--------|
| All Docker phases | SKIPPED |

Docker phases skipped: ports `3052`, `5433`, and `8081` are already in use on the host (existing dev stack). Spinning up `pr-review-385`-prefixed compose project would have collided.

### E2E Browser Tests

| Project | Status |
|---------|--------|
| All E2E | SKIPPED — Docker prerequisite skipped |

Author has not yet driven the bench against the dev stack to capture real four-size numbers; PR description acknowledges this.

---

### Re-review of CRITICAL fixes from previous pass

**CRITICAL #1 — rAF wrapper accumulation.** **CONFIRMED FIXED.** `installBenchInit` is now hoisted out of the per-size loop and called exactly once at line 617 in `main()`. Inside the init script, the counter reset at line 387 (`w.__bench = {...}`) runs on every navigation, while the rAF wrapping is gated by `__benchInstalled` (lines 388-398) so it can't double-wrap within a single page lifetime. Each `page.goto` discards `window`, so `__benchInstalled` is `undefined` on the new page and the wrapper is reinstalled exactly once per navigation. Verified the Playwright `addInitScript` semantics against official docs.

**CRITICAL #2 — URL routing for global graph.** **CONFIRMED FIXED.** Bench now navigates to `/graph?full=1` (line 426) — no `focus=` param. Cross-checked against `frontend/src/features/graph/GraphPage.tsx:136`: `wantsGlobal = !focusPageId && showFullGraph` plus `initialShowFullGraph = searchParams.get('full') === '1'` (line 98). With `focus=` absent and `full=1` present, the page hits the global query (`useGraphData`) calling `/api/pages/graph`, which is the path that returns the seeded corpus.

**CRITICAL #3 — RBAC table for bench user.** **CONFIRMED FIXED.** `grantSpaceAccess` (lines 188-195) now inserts into `space_role_assignments` with the `editor` role. Verified that:
- `space_role_assignments` is the exact table `getUserAccessibleSpaces` queries (`backend/src/core/services/rbac-service.ts:298-308`).
- The `editor` role is seeded by migration 039 (`backend/src/core/db/migrations/039_rbac.sql:21`) with `read,comment,edit,delete` permissions — read access to the space is sufficient for the graph endpoint.
- The unique constraint `(space_key, principal_type, principal_id)` exists (line 52 of 039_rbac.sql) so the `ON CONFLICT` clause is valid.
- `cleanupSize` deletes the assignment first (line 202), correctly handling the no-cascade lifecycle.

### New finding — bench measures the UNION of all seeded sizes, not each size in isolation

A concrete bug introduced by the routing fix in CRITICAL #2: the new URL `/graph?full=1` does NOT include any `space=` filter, so the global query is sent without a `spaceKey` param. The backend (`backend/src/routes/knowledge/pages-embeddings.ts:111-115`) then returns nodes from the **full RBAC-accessible set**.

`main()` seeds ALL requested sizes upfront in the loop at lines 600-606, granting the bench user editor access to every `PERF-BENCH-<size>` space, BEFORE the browser launches. So when `runOne` measures size 500, the user already has access to `PERF-BENCH-500 + 1000 + 2000 + 5000` and the rendered graph contains all 8500 nodes. The same union is rendered for every iteration.

Concrete consequences:
- The four "size" samples in `perf/graph-bench-results.json` are not actually four sizes — they are four measurements of the same union graph (or the largest seeded so far, depending on order).
- The issue's decision rule is anchored on the **2000-node** sample (`decide()` at line 549; tests at line 92-101). With this bug, the "2000 sample" actually renders ~8500 nodes. fps/convergence will look worse than they actually are at 2000, biasing the verdict toward "Phase 2 justified".
- The `samplePageId` field claims to identify a stable handle from the seeded corpus, but each iteration sees a graph spanning all four spaces, so a single id from the 2000 fixture is not a meaningful "stable handle" for that sample.

Suggested fix (any of):
1. Pass `&space=PERF-BENCH-${size}` in the URL so the global query gets `?spaceKey=PERF-BENCH-<size>` and the backend filters correctly. From `GraphPage.tsx:90-91` and `useGraphData` in `graph-hooks.ts:88-92`, this works end-to-end.
2. Seed/clean per-iteration: cleanup all sizes on entry, then for each size seed-then-measure-then-clean before the next size.
3. Per-iteration RBAC: seed everything upfront but only grant the bench user access to ONE space at a time before each runOne; revoke before the next.

Option (1) is the smallest change and matches what a real user would do (filter to one space). It also exercises the same code path the GraphPage's space-filter UI uses.

### Other code review findings

| 1 | CRITICAL | `scripts/perf-graph-bench.ts:426` | Bench navigates to `/graph?full=1` with no `space=` filter; with all sizes seeded simultaneously the global graph endpoint returns the union of all `PERF-BENCH-*` spaces every iteration. See "New finding" section above. |
| 2 | INFO | `scripts/perf-graph-bench.ts:413-416` | Pre-`page.goto` `page.evaluate` reset of `__bench` is dead code. The init script at line 387 already resets counters on every navigation, and `goto` always replaces `window`, so the values evaluate sets get discarded before they could be observed. The accompanying comment ("so the values are clean even if `goto` is fast enough to start the SPA before the init runs") describes a scenario that can't occur — `addInitScript` is documented to run before any document scripts. Harmless but misleading. |
| 3 | INFO | `scripts/perf-graph-bench.ts:529` | `samplePageId` is retained on the JSON output but no consumer ever reads it back. Either drop it or document what downstream artefact will use it. |
| 4 | INFO | `scripts/perf-graph-bench.ts` (no test) | `installBenchInit`'s "install once" guarantee is invariant-critical and has no automated coverage. Hard to test without booting Playwright, but a smoke test that navigates twice and asserts `window.requestAnimationFrame !== <wrapped twice>` would catch a regression if someone moves the call back into the loop. Acceptable to defer. |
| 5 | INFO | `scripts/perf-graph-bench.ts:611-612` | The bench uses Chromium's default `prefers-reduced-motion: no-preference`. If a future CI runner forces `prefers-reduced-motion: reduce`, `GraphPage.tsx:549` short-circuits `cooldownTicks=0` and convergence reads as ~0 ms. Worth either explicitly setting `reducedMotion: 'no-preference'` on `newContext` or noting this in the README's caveats. |
| 6 | INFO | `backend/src/core/services/rbac-service.ts:281` (out-of-PR cache TTL) | The bench writes RBAC assignments directly to Postgres without invalidating the Redis cache (`getUserAccessibleSpaces` caches for 60s). On a second consecutive run within the cache window, a stale `[]` cached from before the first seed could deny access. Currently safe because the first run seeds before any graph request and the cache key is per-user; flag for awareness if the bench gains a concurrent-runner mode. |

### Acceptance Criteria (Issue #380)

| AC-1 | Phase 1 benchmark lands with numbers in the PR comment | NOT YET | Author acknowledges; numbers will follow once a clean run completes. The bench cannot produce trustworthy numbers until the union-of-spaces issue is addressed. |
| AC-2 | Phase 2 (off-thread layout) — only if Phase 1 says so | N/A this PR | Deferred to a follow-up. |
| AC-3 | Phase 3 (renderer swap) — only if Phase 1 says so | N/A this PR | Deferred. |

### Story Validation (Issue #380)

**Issue intent**: Build a synthetic 500/1000/2000/5000-node graph benchmark whose **measured** numbers determine whether Phase 2 (server-side ForceAtlas2) is worth doing. The issue is explicit that paste-able numbers, not just a verdict, are the deliverable.

**Story completion**: PARTIALLY ADDRESSED.

The benchmark *script* is now structurally close — three of four CRITICAL bugs from the prior review are genuinely resolved. But the script as it stands still cannot produce per-size numbers because the global graph endpoint sees all four seeded spaces at once. Until the bench filters to one space per iteration (or seeds one at a time), the four "samples" will all describe the same union graph. The issue's decision rule (anchored on the 2000-node sample) cannot be applied meaningfully against a union sample.

**User journey**: Before: no measurement -> can't make data-driven Phase 2 decision. After (with this PR merged as-is): `npx tsx scripts/perf-graph-bench.ts` runs to completion and writes a JSON report — but the four `samples[]` entries describe `pageCount=500/1000/2000/5000` while the rendered graph in each iteration is the union (~8500 nodes when running the default sweep). The verdict is misleading.

**Gaps**:
- Per-size isolation in the rendered graph (the new CRITICAL above).
- End-to-end real-numbers run still pending — author has flagged this and it's reasonable to land Phase 1 in two steps, but the union bug should be fixed first so the run is meaningful.

**Evidence**: bench navigates to `/graph?full=1` (no `space=`) at `scripts/perf-graph-bench.ts:426`; GraphPage routes to global query without filter (`frontend/src/features/graph/GraphPage.tsx:136-137`); backend returns union of accessible spaces (`backend/src/routes/knowledge/pages-embeddings.ts:111-115`); all sizes seeded before any measurement (`scripts/perf-graph-bench.ts:600-606`).

### Suggested Actions

1. **[REQUIRED]** Fix the union-of-spaces measurement bug. Smallest change: navigate to `/graph?full=1&space=${seed.spaceKey}` so each iteration filters to its own `PERF-BENCH-<size>` space. This also makes `samplePageId` meaningful again.
2. **[CONSIDER]** Drop the dead pre-`goto` `page.evaluate` reset (lines 413-416) and the misleading comment, OR keep the reset but rewrite the comment to describe its real role (defence-in-depth for re-running `runOne` on a non-fresh page).
3. **[CONSIDER]** Set `reducedMotion: 'no-preference'` on `browser.newContext({ ... })` to insulate the convergence measurement from a future CI runner that defaults to reduced motion.
4. **[OPTIONAL]** Capture and paste the four real samples + verdict in the PR before merging — the issue treats the numbers as the deliverable, not the script.

### Verdict: REQUEST_CHANGES

**Reason**: The three CRITICAL fixes from the prior review do all hold up under scrutiny — `installBenchInit` is now correctly single-shot, `?full=1` (no `focus=`) does route to the global graph, and `space_role_assignments` with the seeded `editor` role does grant the access the graph endpoint actually checks. However, the routing fix exposed a separate bug: with no `space=` filter and all four sizes seeded upfront, every iteration measures the union of all `PERF-BENCH-*` spaces (~8500 nodes), not the per-size graph. That makes the bench's four samples non-comparable and renders the issue's 2000-node decision rule meaningless. This is a genuine correctness bug introduced (or surfaced) by the routing fix, so the PR is one small change away from approval but isn't yet there.

---
Reviewed by PR Reviewer Agent (re-review of 40bf559) | 2026-04-28

laboef1900:
Addressed the recheck findings in `5489763`.

**CRITICAL — per-iteration space filter (the new finding from the recheck)**

The bench now navigates to `/graph?full=1&space=PERF-BENCH-${size}`. `useGraphData` forwards the `space=` URL param as `spaceKey=…` to `/api/pages/graph` (graph-hooks.ts:91), which the backend uses to filter to the requested space. Each iteration now measures only its own seeded corpus, not the merged ~8500-node union across all four sizes.

**INFO items also addressed**

- Removed the dead pre-`goto` rAF counter reset. The `addInitScript` body already resets `__bench` on every navigation, so the prior `page.evaluate` was a no-op against the previous page's window.
- Pinned `reducedMotion: 'no-preference'` on the Playwright context. GraphPage uses `cooldownTicks={prefersReducedMotion ? 0 : 100}` — without this pin, a host-OS `prefers-reduced-motion: reduce` would short-circuit the simulation and convergence would bottom out at 0 ms regardless of size.
- Extracted the init-script body into an exported `benchInitScriptBody`. Three new vitest cases pin the single-shot guarantee in a fake-DOM: rAF wrapped exactly once across replays, counters reset every replay, auth token seeded into localStorage. A future regression of the addInitScript pattern can't slip past the unit lane now.
- Added a best-effort RBAC cache flush. `getUserAccessibleSpaces` caches `rbac:spaces:<userId>` for 60 s — consecutive runs with different `--sizes` could read a stale space list. When `REDIS_URL` is set, the script flushes the bench user's RBAC keys after seeding; falls back to a printed warning otherwise. README documents the caveat + manual workaround.

The `samplePageId` field in the JSON output was kept intentionally — it's a debugging hook for a sample row in the seeded space if a result looks weird, and the field is documented in the `BenchSample` interface.

**Verification**

- `npx vitest run --config scripts/vitest.config.ts` — 14/14 pass (3 new for the single-shot guarantee)
- `npx tsc --noEmit` clean on the script + test
- `npx tsx scripts/perf-graph-bench.ts` parses cleanly (only fails on Postgres ECONNREFUSED, expected when Postgres is on the internal-only data-net)

laboef1900 (commented):
## Automated PR Review (Third Pass)

**PR**: #385 -- feat(#380 Phase 1): synthetic graph perf benchmark
**Branch**: feature/380-graph-perf-bench -> dev
**Diff size**: 1095 lines (+1069/-0)
**Reviewed at**: 2026-04-28

---

### Static Checks

| Check | Status | Details |
|-------|--------|---------|
| TypeScript | PASS | clean across contracts/backend/frontend |
| Lint | PASS | one upstream `eslint-plugin-boundaries` deprecation warning, unrelated to PR |
| Bench unit tests | PASS | 14/14 (`npx vitest run --config scripts/vitest.config.ts`) |
| Backend Tests | SKIPPED | not re-run in this pass; CE base unchanged from PR #378 |
| Frontend Tests | SKIPPED | not re-run in this pass; CE base unchanged |
| Production Build | SKIPPED | scripts/* is outside the workspace build graph |
| npm audit | SKIPPED | this pass focused on the targeted fixes |
| Branch Target | PASS | feature/* -> dev (matches CLAUDE.md Git Workflow) |

### Docker App Test

| Check | Status |
|-------|--------|
| All Docker phases | SKIPPED |

Docker tests skipped: ports 5433/3052 held by user dev stack, consistent with prior reviews. Verdict capped at COMMENT regardless of code-review outcome.

### E2E Browser Tests

| Project | Status |
|---------|--------|
| auth/setup/chromium | SKIPPED |

E2E tests skipped: dependent on Docker stack.

### Verification of Each Targeted Fix

| Concern | Result |
|---------|--------|
| Per-iteration `&space=` filter reaches backend | VERIFIED end-to-end |
| `benchInitScriptBody` single-shot test | VERIFIED, real contract pinned |
| `reducedMotion: 'no-preference'` Playwright option | VERIFIED, but defensive (already the default) |
| `flushRbacCache` covers right keys | VERIFIED for the actual code path |
| Dead pre-goto `page.evaluate` removed | CONFIRMED removed at runOne entry |

#### 1. Per-iteration space filter — verified

Traced the URL param end-to-end:

- `scripts/perf-graph-bench.ts:464` builds `?full=1&space=PERF-BENCH-${size}`.
- `frontend/src/features/graph/GraphPage.tsx:90` parses `searchParams.get('space')` into `filterSpaces` (split-on-comma, filtered).
- `frontend/src/features/graph/GraphPage.tsx:137` passes `filterSpaces` as the second argument to `useGraphData(viewMode, filterSpaces, wantsGlobal)`.
- `frontend/src/features/graph/graph-hooks.ts:91` joins the array on `,` and encodes as `spaceKey=...` on the wire.
- `backend/src/routes/knowledge/pages-embeddings.ts:93` parses `spaceKey` via `SpaceKeysSchema` (re-splits on `,`, trims, drops empties).
- `backend/src/routes/knowledge/pages-embeddings.ts:112-115` intersects `filterSpaceKeys` with `getUserAccessibleSpaces(userId)` to produce `effectiveSpaces`.
- `backend/src/routes/knowledge/pages-embeddings.ts:144` SQL filters `WHERE cp.space_key = ANY($1::text[])`.

So when the bench runs `?space=PERF-BENCH-500`, only that one space's pages are returned regardless of how many other PERF-BENCH-* spaces the user has access to. The size sweep is correctly scoped.

Cache key (line 101) includes the sorted space-key set, so different filter selections do not collide in Redis even when the same user toggles between sizes.

#### 2. `benchInitScriptBody` single-shot test — honestly exercises the contract

Walked through `runScript` (test:138-148):
- It serialises `benchInitScriptBody.toString()` and wraps it in `new Function('window', 'localStorage', 'performance', '(${...})({token: "tok"});')`.
- Inside the function body, `window`, `localStorage`, `performance` resolve to the parameters (which is the fake `windowProxy`).
- The body's `w === window === windowProxy` so `w.__benchInstalled = true` actually sets a property on the proxy that survives across calls.

The "exactly once" assertion at line 163 (`expect(windowProxy.requestAnimationFrame).toBe(afterFirst)`) holds only if the second/third runs hit the `if (w.__benchInstalled) return` early-out before the rAF replacement. If a future regression removed that guard, the second `runScript` would replace `requestAnimationFrame` with a *new* wrapper closure (different identity) and the assertion would fail.

The test honestly pins the contract. No hole.

The reset-on-replay test (170-183) and the localStorage seed test (185-196) cover the other two responsibilities of the script. `__benchInstalled` is verified to be `true` after the first install (line 159), confirming the guard flag is set.

#### 3. `reducedMotion: 'no-preference'` — correct option but defensive

Cross-checked Playwright docs (api/params.md, `context-option-reducedMotion`):

> Emulates `'prefers-reduced-motion'` media feature, supported values are `'reduce'`, `'no-preference'`. ... **Defaults to `'no-preference'`**.

So `reducedMotion: 'no-preference'` is the correct API and the right value for emulating "user has motion enabled" — when set, `window.matchMedia('(prefers-reduced-motion: reduce)').matches` returns `false`, which is what the GraphPage's `useReducedMotion` hook reads (`GraphPage.tsx:152-155`).

However: this is **already the Playwright default**. Playwright contexts emulate media queries — they do NOT pick up the host OS `prefers-reduced-motion` setting unless explicitly told to. The PR comment ("If the host's `prefers-reduced-motion: reduce` setting leaks through ...") slightly overstates the danger. The pin is harmless and self-documenting, just not load-bearing.

This is INFO, not a defect.

#### 4. `flushRbacCache` — covers the keys that matter for the bench

Cross-checked against `backend/src/core/services/rbac-service.ts`:

| Cache key | Defined at | Flushed by bench? | Used by graph endpoint? |
|-----------|-----------|-------------------|-------------------------|
| `rbac:perms:<userId>:<spaceKey>` | rbac-service.ts:11 | NO | NO (graph endpoint never calls `userHasPermission`) |
| `rbac:spaces:<userId>` | rbac-service.ts:15 | YES | YES (load-bearing — `getUserAccessibleSpaces`) |
| `rbac:admin:<userId>` | rbac-service.ts:19 | YES | indirectly (via `isSystemAdmin` inside `getUserAccessibleSpaces`) |
| `rbac:global:<userId>` | rbac-service.ts:23 | YES | NO |

`grep`-confirmed: `pages-embeddings.ts` only imports `getUserAccessibleSpaces` from rbac-service (line 6), and only that function is called in the graph handler (lines 111, 260). So the only flush key that's load-bearing for the bench is `rbac:spaces:<userId>`, which IS flushed.

The other two flushes are defensive and free. The missing `rbac:perms:*` key is irrelevant for this code path and would, in any case, require enumerating spaces. Not a bug.

#### 5. Pre-goto reset removed — confirmed

`runOne` (scripts/perf-graph-bench.ts:443-466) no longer contains the dead `page.evaluate` reset. The init-script body sets `w.__bench = {...}` on every replay (line 422), so each `goto` clears the counters. Mid-iteration reset (lines 515-520, before the FPS measurement window) is preserved — that's a different reset, intentional and correct.

### Code Review Findings

| 1 | INFO | scripts/perf-graph-bench.ts:655-661 | The comment for `reducedMotion: 'no-preference'` overstates necessity. Playwright already defaults to `'no-preference'` (per official docs). The pin is defensive/self-documenting; the host OS setting does not leak into Playwright contexts. Consider rewording to "pinned for explicitness; matches Playwright default" so future readers don't infer that omitting the line would actually break the bench. |
| 2 | INFO | scripts/perf-graph-bench.ts:354 | `flushRbacCache` deletes `rbac:admin:<userId>` and `rbac:global:<userId>` defensively, but the graph endpoint only consumes `rbac:spaces:<userId>` (and indirectly the admin cache via `isSystemAdmin` inside `getUserAccessibleSpaces`). A short comment noting that `rbac:perms:<userId>:<spaceKey>` is intentionally NOT flushed — because the graph endpoint never calls `userHasPermission` — would prevent a future contributor from "completing the set" with a SCAN/MATCH that costs Redis CPU on every run. |
| 3 | INFO | scripts/perf-graph-bench.ts:728 | The `import.meta.url === \`file://${process.argv[1]}\`` test is platform-fragile (Windows uses `file:///C:/...`). The `process.argv[1]?.endsWith('perf-graph-bench.ts')` fallback covers the common case, but a Windows POSIX-mode user running `npx tsx` could hit a `file:///` vs `file://` mismatch. Low-impact (bench is dev-machine tooling, the README assumes a Linux/macOS docker compose stack), so noting only. |
| 4 | INFO | perf/README.md:284 | The README mentions flushing `compendiq:cache:<userId>:spaces:list`, but the actual key is `rbac:spaces:<userId>` (per rbac-service.ts:15). The advice given to manually flush is correct in spirit but the key name in the README doesn't match the source. Either align the doc to `rbac:spaces:<userId>` or use `KEYS rbac:*:<userId>*` as the flush pattern. |
| 5 | INFO | scripts/perf-graph-bench.ts | The PR description acknowledges the bench has not been driven against the dev stack with real numbers yet. That is the actual deliverable for #380 ("decide whether Phase 2 is justified"). Approving the harness is fine; the verdict for #380 itself remains "evidence pending". |

### Acceptance Criteria (Issue #380)

Issue #380 is a "decide whether Phase 2 is needed" investigation. Phase 1 of #380 is "build a synthetic-corpus benchmark and run it". The PR explicitly delivers Phase 1 (the harness) and defers the actual measurement to a follow-up comment.

| AC-1 | Synthetic graph generator at multiple sizes (500/1000/2000/5000) | PASS | scripts/perf-graph-bench.ts:79-82, seedSize:211 |
| AC-2 | Realistic edge density (top-K=5, score >= 0.6) | PASS | EDGES_PER_NODE=5 (line 100), MIN_SCORE=0.6 (line 101) |
| AC-3 | Measurements: TTFP, layout convergence, FPS, JS heap | PASS | runOne returns BenchSample with all four (line 312-331) |
| AC-4 | Decision rule applied (fps<30 OR convergence>5000ms at 2000 nodes) | PASS | decide() at line 587, 2000-node anchored |
| AC-5 | Cleanup of synthetic data | PASS | cleanupSize:199, called from finally block:687-688 |
| AC-6 | Run against dev stack and produce numbers | DEFERRED | acknowledged in PR description, follow-up comment |

### Story Validation (Issue #380)

**Issue intent**: Decide whether the graph rendering needs a Phase 2 (canvas/regl rewrite, server-side clustering, etc.) by measuring the current implementation under realistic load.

**Story completion**: PARTIALLY ADDRESSED — the harness is fully delivered, but the decision itself depends on running the harness, which has not yet happened. The PR description says numbers will follow in a comment.

**User journey**: Before: maintainers had no way to measure graph perf at scale without a real Confluence corpus. After: anyone can run `npx tsx scripts/perf-graph-bench.ts` against a dev stack and get a JSON report with a verdict.

**Gaps**: No actual run-on-dev numbers attached. This is acknowledged. Once the numbers land, AC-6 closes and #380 can move to a follow-up issue (Phase 2 justified or not).

**Evidence**: `scripts/perf-graph-bench.ts` is a complete harness; `scripts/perf-graph-bench.test.ts` pins 14 unit tests; `perf/README.md` documents usage and known gotchas.

### Resolution of Prior Review Findings

| # | Severity in Round 2 | File:Line | Resolution |
|---|---------------------|-----------|------------|
| 1 | CRITICAL | scripts/perf-graph-bench.ts (URL building) | RESOLVED — `&space=...` per-iteration filter; verified end-to-end |
| 2 | INFO | dead pre-goto `page.evaluate` reset | RESOLVED — removed |
| 3 | INFO | reduced-motion insulation | RESOLVED (defensively) |
| 4 | INFO | no test for single-shot guarantee | RESOLVED — 3 new vitest cases pin the contract |
| 5 | INFO | RBAC cache TTL gotcha | RESOLVED — `flushRbacCache` + README documents fallback |
| 6 | INFO | unused `samplePageId` | INTENTIONALLY KEPT — documented on `BenchSample` interface as a debugging hook |

All Round 2 findings are addressed. No new CRITICAL or WARNING findings in this pass.

### Suggested Actions

1. **[OPTIONAL]** Reword the `reducedMotion` comment (scripts/perf-graph-bench.ts:655-661) so it doesn't imply omitting the option would break the bench. Playwright defaults to `'no-preference'`; the explicit pin is for clarity, not necessity.
2. **[OPTIONAL]** Add a one-line comment near `flushRbacCache` explaining why `rbac:perms:*` is intentionally NOT flushed (graph endpoint never calls `userHasPermission`). This will save a future contributor from over-engineering the flush.
3. **[OPTIONAL]** Align `perf/README.md` "manual workaround" key name to the actual `rbac:spaces:<userId>` key from rbac-service.ts:15. Current mention of `compendiq:cache:<userId>:spaces:list` does not match the codebase.
4. **[FOLLOW-UP]** Run the bench against the dev stack and post the JSON results (or a summary table) on this PR before merging the Phase 1 conclusion into a Phase 2 plan. The harness is approvable; the issue's decision still needs the numbers.

### Verdict: COMMENT

**Reason**: All three Round 2 CRITICAL fixes verified end-to-end. The new single-shot test honestly pins the contract (14/14 pass). RBAC flush keys cover the load-bearing key (`rbac:spaces:<userId>`); the other two flushes are defensive but free. `reducedMotion` option is correct (and is also Playwright's default — the comment overstates necessity). No new CRITICAL or WARNING findings. Verdict capped at COMMENT because Docker phases were skipped (port conflict, consistent with prior reviews) and the actual bench numbers from a dev-stack run are still outstanding (acknowledged in PR description).

The harness is solid and ready to merge. The Phase 2 justification decision (the actual point of #380) is awaiting the bench run.

---

laboef1900:
## Phase 1 numbers (first run)

Driven against the local dev stack on this machine (linux/aarch64 — Playwright using its arm64 fallback build, so absolute numbers are conservative vs a "mid-tier laptop" running the official x64 build). Postgres + Redis port-forwarded via socat sidecars on `compendiq_data-net`; bench seeded against the running `compendiq-backend-1` (`:3052`) + `compendiq-frontend-1` (`:8081`). Bench user is the freshly-registered `benchuser` with `editor` role granted on each `PERF-BENCH-<size>` space.

| Size | TTFP (ms) | Layout convergence (ms) | Drag fps | Heap (MB) | Edges |
|---|---|---|---|---|---|
|  500 |  504 |  2 791 |  74.5 | 48.1 | 1 500 |
| 1000 | 1 120 | 31 113 ⚠ |  82.5 | 48.1 | 3 000 |
| 2000 |   928 | 30 959 ⚠ |  90.0 | 48.1 | 6 000 |
| 5000 | 1 284 |  1 357 | 136.5 | 48.1 | 15 000 |

**TTFP** scales reasonably with corpus size. **Drag fps** is comfortably above 60 fps (well above the issue's 30 fps threshold) at every size — including 5000 nodes. **Heap** is invariant at 48.1 MB across sizes; that's `performance.memory` quantisation (Chrome rounds to ~10 MB increments and only updates lazily), not a real signal.

**Convergence detector is still flaky.** It flips between 1.4 s (5000) and the 30 s timeout (1000, 2000) — non-monotonic in size, which means the rate-based detector in `bfb1a8f` (now an improvement over the original "rAF goes fully quiet" approach that was always pegged at the timeout) still can't reliably distinguish "force simulation stopped" from "background rAF activity from Framer Motion / Radix transitions". The convergence column is unsafe to base a Phase-2 decision on in this run.

**Reading by the fps criterion alone** — which IS reliable — Phase 2 is **NOT justified** at any size up to 5000 nodes. The issue's threshold is "<30 fps drag at 2000 nodes" and we measure 90 fps at 2000.

## Recommended follow-up before applying the issue's decision rule

Before flipping the Phase 2 switch (server-side ForceAtlas2 + `page_graph_layout` + sigma.js evaluation), the convergence signal needs to be load-bearing. The cleanest path is **GraphPage-side instrumentation**: when `?full=1` is present, expose `window.__compendiqGraph = { onEngineStop: () => …, settled: false }` from the GraphPage so the bench can observe the simulation's actual stop event instead of inferring it from rAF rate. That's a small frontend change and would replace the rate-based heuristic with the exact signal the issue calls for ("when `cooldownTicks` hits 0").

Tracking that as a Phase-1.5 follow-up. The bench framework, fixtures, RBAC seeding, ttfp / fps / interaction harness, and the four-size sweep all work; only the convergence detector needs the cleaner upstream signal.

## Caveat / reproducibility

- Postgres on dev compose is internal-only — needed two `socat` port-forwarders on `compendiq_data-net` to expose `:5440 → postgres:5432` and `:6380 → redis:6379`. README documents the workaround.
- The arm64 fallback Chromium is slower than x64 official builds — fps numbers under-state the headroom on a typical x64 dev machine.
- `samplePageId` values in the table differ from the run because consecutive seeds reuse `id` sequence; the "sample" id is just an internal handle.

JSON output (`perf/graph-bench-results.json`, gitignored under `perf/*.json`) preserved locally for reference.

PR #650, opened 2026-05-14
refactor(security): use constant format string in FeatureErrorBoundary logger

### LINKED ISSUE #640: security(semgrep): unsafe-formatstring in FeatureErrorBoundary.tsx
## Semgrep finding
- **Rule:** `javascript.lang.security.audit.unsafe-formatstring`
- **Severity:** low (confidence: low)
- **CWE:** CWE-134 (Use of Externally-Controlled Format String)
- **OWASP:** A03:2021 - Injection

## Locations
- `frontend/src/shared/components/feedback/FeatureErrorBoundary.tsx:40` — `console.error(`[FeatureErrorBoundary] ${this.props.featureName} crashed:`, error, errorInfo);`

## Analysis

**False positive — template literal, not a `util.format`-style format string, and the interpolated value is developer-controlled.**

The Semgrep rule targets cases where an externally-controlled string is passed as the first argument to a function that interprets `%s` / `%d` / `%o` placeholders (like `util.format`, `printf`, or in some browsers `console.log`). Two reasons this match is benign:

1. **It is a JavaScript template literal (`` ` ``…`${...}`…`` ` ``), not a format-specifier string.** Interpolation has already been resolved by the JS engine *before* `console.error` ever sees the string. Any `%s` characters that happened to be inside `featureName` would simply be printed verbatim — there is no second pass of format-specifier interpretation. (`console.error` in V8 does interpret `%s` only when the **first argument is a literal string containing format specifiers**; the engine still receives the already-interpolated final string here.)

2. **`featureName` is a developer-supplied React prop, not user input.** Grep across `frontend/src/` confirms callers pass static literals like `"ArticleView"`, `"SidebarTree"`, etc. There is no user-controlled data path into this prop. Even if a future caller passed something dynamic, point (1) still neutralizes the format-string risk; the worst case is a malformed log line.

3. The line runs in the **browser**, so even classical format-string memory-corruption (the original CWE-134 scenario in C) is not applicable — the JS log function is a thin DOM wrapper.

## Proposed remediation

Suppress with `nosemgrep` plus a one-line justification. No behavior change.

```tsx
componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
  // nosemgrep: javascript.lang.security.audit.unsafe-formatstring
  // Template literal (already-interpolated) + developer-controlled featureName prop.
  console.error(
    `[FeatureErrorBoundary] ${this.props.featureName} crashed:`,
    error,
    errorInfo,
  );
}
```

Alternative (slightly preferred for log-aggregation friendliness) — pass `featureName` as a structured second arg so the message becomes a constant string:

```tsx
console.error('[FeatureErrorBoundary] crashed', {
  featureName: this.props.featureName,
  error,
  errorInfo,
});
```

This makes the format string a literal constant *and* gives Sentry / browser dev tools a structured object to expand.

## References
- Semgrep rule: https://semgrep.dev/r?q=javascript.lang.security.audit.unsafe-formatstring
- CWE-134: https://cwe.mitre.org/data/definitions/134.html
- MDN template literals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
- Chrome DevTools `console.log` format specifiers: https://developer.chrome.com/docs/devtools/console/format-style#format-specifiers

---
_Filed by automated Semgrep triage. Source finding: Semgrep cloud, deployment `simons_personal_org`._

laboef1900:
## Summary

Closes #640. `FeatureErrorBoundary.componentDidCatch` previously passed a template literal as the format-position argument to `console.error`, with the feature name interpolated into the string. Any feature whose `featureName` prop contained `%s` / `%d` / `%o` would have those specifiers consumed by `console.error`'s util.format-style formatter, allowing log forgery (semgrep `javascript.lang.security.audit.unsafe-formatstring.unsafe-formatstring`, CWE-117 / CWE-134).

The format string is now a constant; the feature name is passed as a positional `%s` argument.

## Before / after

```diff
  componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
    console.error(
-     \`[FeatureErrorBoundary] \${this.props.featureName} crashed:\`,
+     '[FeatureErrorBoundary] %s crashed:',
+     this.props.featureName,
      error,
      errorInfo,
    );
  }
```

The matching test was updated to assert against the new argument layout (`call[0]` is the constant format string, `call[1]` is the feature name).

## Scope check

`grep -rn 'console\.(error|log|warn)(\`' frontend/src/shared/components/feedback frontend/src/shared/lib frontend/src/shared/hooks` — no other instances of the same anti-pattern in the immediate scope. A broader `semgrep --config=p/javascript` over those three directories also returns 0 unsafe-formatstring findings, so nothing else needs the same fix in this PR.

## Semgrep rerun

Pre-change:

```
Findings: 1 (1 blocking)
  javascript.lang.security.audit.unsafe-formatstring.unsafe-formatstring
    40┆ \`[FeatureErrorBoundary] \${this.props.featureName} crashed:\`,
```

Post-change on the same file:

```
Findings: 0 (0 blocking)
Rules run: 378
Targets scanned: 1
```

## Verification

- `npm run lint -w frontend` — clean
- `npm run typecheck -w frontend` — clean
- `npm test -w frontend -- FeatureErrorBoundary` — **10/10 passing**
- Semgrep rerun on `FeatureErrorBoundary.tsx` — 0 findings

## Test plan

- [x] Unit test still asserts that our specific componentDidCatch log was emitted, now matching the format-string layout.
- [x] Lint, typecheck, semgrep all pass.
- [ ] Manual smoke: render a feature that throws — verify the dev-tools console shows the boundary's `[FeatureErrorBoundary] <name> crashed:` line plus the React error and ErrorInfo (jsdom test already covers this).

### DIFF
--- a/frontend/src/shared/components/feedback/FeatureErrorBoundary.test.tsx
+++ b/frontend/src/shared/components/feedback/FeatureErrorBoundary.test.tsx
@@ -132,11 +132,14 @@ describe('FeatureErrorBoundary', () => {
     );
 
     // React itself calls console.error, and our componentDidCatch also logs.
-    // Verify our specific log message was included.
+    // Verify our specific log was included. The boundary now uses a constant
+    // format string with `%s` interpolation (the feature name is passed as
+    // the second argument) — see #640 / unsafe-formatstring lint rule.
     const ourLog = consoleSpy.mock.calls.find(
       (call) =>
         typeof call[0] === 'string' &&
-        call[0].includes('[FeatureErrorBoundary] Article Viewer crashed:'),
+        call[0] === '[FeatureErrorBoundary] %s crashed:' &&
+        call[1] === 'Article Viewer',
     );
     expect(ourLog).toBeDefined();
   });
--- a/frontend/src/shared/components/feedback/FeatureErrorBoundary.tsx
+++ b/frontend/src/shared/components/feedback/FeatureErrorBoundary.tsx
@@ -37,7 +37,8 @@ export class FeatureErrorBoundary extends Component<
 
   componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
     console.error(
-      `[FeatureErrorBoundary] ${this.props.featureName} crashed:`,
+      '[FeatureErrorBoundary] %s crashed:',
+      this.props.featureName,
       error,
       errorInfo,
     );


laboef1900 (commented):
## Independent review

### Verdict
Approve

### Security correctness
Better than a suppression. The original Semgrep finding was arguably a false positive (template literal is already-interpolated before `console.error` sees it, and `featureName` is a developer-supplied prop), but the PR opted for the **conservative real fix**: a constant format string with `%s` and the feature name as a positional arg. That's the textbook CWE-117/CWE-134 mitigation pattern and it eliminates the rule match entirely with no `nosemgrep` debt. Browser `console.error` does interpret `%s` in literal first-arg strings, so the new form is also semantically equivalent in dev tools.

### Test coverage
Adequate. The test was correctly updated to assert against `call[0] === '[FeatureErrorBoundary] %s crashed:' && call[1] === 'Article Viewer'` instead of `call[0].includes(...)`. That tightens the assertion (it now pins the exact format-string layout) rather than weakening it.

### Concerns
None.

### Nits (won't block)
- Worth a follow-up to grep the rest of `frontend/src/` for the same template-string-in-console pattern. The PR body says it did this for `feedback/`, `lib/`, `hooks/` and found nothing, but the broader app surface (e.g. `features/*/`) wasn't covered. Low priority — most other `console.error` sites use the structured form already.
- Issue #640 suggested an alternative structured-arg form (`console.error('[FeatureErrorBoundary] crashed', { featureName, error, errorInfo })`) which would be slightly more aggregator-friendly. Not a blocker; the current form is fine and changes less.

PR #648, opened 2026-05-14
refactor(security): centralize unsafe TLS bypass into single audited helper

laboef1900:
## Summary

Closes #634.

Both `tls-config.ts` and `llm-config.ts` previously inlined `return { rejectUnauthorized: false };` when their respective `*_VERIFY_SSL=false` escape hatches were active, which tripped semgrep `problem-based-packs.insecure-transport.js-node.bypass-tls-verification` at two call sites.

The bypass itself is intentional (documented in `CLAUDE.md` / `.env.example` for on-prem Confluence DC + self-hosted LLM proxies with self-signed corporate CAs), gated behind env vars that default to verification-on, and accompanied by a runtime warning when disabled. Rather than annotate every call site with `nosemgrep`, this PR extracts a single helper `unsafeDisableTlsVerification()` in `backend/src/core/utils/unsafe-tls.ts`. The helper holds the **only** `nosemgrep` suppression in the codebase for this rule and a doc comment explaining the gating + the operator opt-in story; both modules now call it.

This makes any future audit of "where do we drop TLS verification?" a single grep for callers of one function.

### Before / After

`backend/src/core/utils/tls-config.ts` (and equivalent in `llm-config.ts`):

```diff
 export function buildConnectOptions(): Record<string, unknown> | undefined {
   if (!verifySsl) {
-    return { rejectUnauthorized: false };
+    return unsafeDisableTlsVerification();
   }
   if (caBundleContents) {
     return { ca: caBundleContents };
   }
   return undefined;
 }
```

New helper (single source of truth, single suppression):

```ts
export function unsafeDisableTlsVerification(): { rejectUnauthorized: false } {
  // nosemgrep: problem-based-packs.insecure-transport.js-node.bypass-tls-verification.bypass-tls-verification -- single, audited usage; gating happens at callsite via env var, runtime warning logged, NODE_EXTRA_CA_CERTS preferred
  return { rejectUnauthorized: false };
}
```

### Behaviour

No change. Same env-var checks, same warnings, same return shape. Existing `tls-config.test.ts` and `llm-config.test.ts` assertions about the `{ rejectUnauthorized: false }` shape still pass unchanged.

### Verification

- `npm run lint -w backend` — clean
- `npm run typecheck -w backend` — clean
- `npx vitest run src/core/utils/unsafe-tls.test.ts src/core/utils/tls-config.test.ts src/core/utils/llm-config.test.ts` — **30/30 pass**
[...]

Issue #634, opened 2026-05-14
security(semgrep): bypass-tls-verification in llm-config.ts & tls-config.ts


laboef1900:
## Semgrep finding
- **Rule:** `problem-based-packs.insecure-transport.js-node.bypass-tls-verification`
- **Severity:** medium (confidence: high)
- **CWE:** CWE-319 (Cleartext Transmission of Sensitive Information)
- **OWASP:** A02:2021 - Cryptographic Failures

## Locations
- `backend/src/core/utils/llm-config.ts:103` — `return { rejectUnauthorized: false };` inside `buildLlmConnectOptions()` when `LLM_VERIFY_SSL=false`.
- `backend/src/core/utils/tls-config.ts:66` — `return { rejectUnauthorized: false };` inside `buildConnectOptions()` when `CONFLUENCE_VERIFY_SSL=false`.

## Analysis

**False positive — intentional opt-in TLS escape hatch.**

Both code paths are gated behind explicit, default-on environment variables (`LLM_VERIFY_SSL` / `CONFLUENCE_VERIFY_SSL`). The user must set them to `false` to disable verification. This is required for on-prem Confluence Data Center and self-hosted LLM proxies that use internal/self-signed CA chains (a very common enterprise configuration).

Existing mitigations already in place:
1. **CA-bundle override is preferred over disabling verification.** Both modules first try `NODE_EXTRA_CA_CERTS` and common system CA paths (`/etc/ssl/certs/ca-certificates.crt`, `/etc/pki/...`). Disabling verification is the last resort.
2. **Runtime warning is logged when verification is disabled:**
   - `backend/src/core/utils/llm-config.ts:77-79` — `logger.warn('LLM_VERIFY_SSL=false ...')`
   - `backend/src/core/utils/tls-config.ts:56-58` — `logger.warn('CONFLUENCE_VERIFY_SSL=false ...')`
3. **Documented in `CLAUDE.md`** under "TLS escape hatches: `LLM_VERIFY_SSL`, `CONFLUENCE_VERIFY_SSL`, `NODE_EXTRA_CA_CERTS`" and in `.env.example`.

## Proposed remediation

Add `// nosemgrep: problem-based-packs.insecure-transport.js-node.bypass-tls-verification` on each `return { rejectUnauthorized: false };` line with a one-line justification comment explaining the env-var gating and the existing warning log. No code/behavior change needed.

Example:
```ts
if (!verifySsl) {
[...]

PR #651, opened 2026-05-14
refactor(security): explicit traversal-safe attachment path helper

### LINKED ISSUE #635: security(semgrep): path-join-resolve-traversal in attachment-handler.ts
## Semgrep finding
- **Rule:** `javascript.lang.security.audit.path-traversal.path-join-resolve-traversal`
- **Severity:** medium (confidence: high)
- **CWE:** CWE-22 (Improper Limitation of a Pathname to a Restricted Directory)
- **OWASP:** A01:2021 - Broken Access Control

## Locations
- `backend/src/domains/confluence/services/attachment-handler.ts:59` — `const safe = path.basename(filename);` followed by `path.join(attachmentDir(pageId), safe)` on line 61.

## Analysis

**False positive — multi-layer path-traversal defenses already in place.**

The flagged `path.join` call (line 61) is already protected on both sides:

1. **`filename` is sanitized via `path.basename()`** (line 59). `path.basename` is the canonical Node.js way to strip directory components: it removes everything up to the final separator, so inputs like `../../../etc/passwd` collapse to `passwd`. Backslashes are also handled on Windows builds. This is the exact mitigation Semgrep's own rule documentation recommends.

2. **`pageId` is sanitized inside `attachmentDir()`** (lines 41-52):
   - Line 43 replaces any `/`, `\\`, `.` with `_` and strips leading/trailing underscores.
   - Lines 47-51 do a `path.resolve()` containment check — if the resolved directory does not start with the resolved `ATTACHMENTS_BASE`, it throws `'Path traversal detected'`.

3. An existing `// nosemgrep` is already on line 60, but Semgrep is now flagging line 59 (the `basename` line) rather than the `path.join` line. The annotation needs to move/expand so Semgrep matches it correctly.

## Proposed remediation

Move/extend the `nosemgrep` so it applies to the line Semgrep actually flags, and add a one-line justification. No behavior change.

Before:
```ts
const safe = path.basename(filename);
// nosemgrep
return path.join(attachmentDir(pageId), safe);
```

After:
```ts
// nosemgrep: javascript.lang.security.audit.path-traversal.path-join-resolve-traversal
// filename is stripped to its basename; pageId is validated by attachmentDir()
// which throws on any path-traversal attempt (see lines 41-52 above).
const safe = path.basename(filename);
return path.join(attachmentDir(pageId), safe);
```

Optionally, add a unit test asserting that `cacheAttachment` rejects/normalizes `../`-style payloads in both `pageId` and `filename` to lock in the invariant.

## References
- Semgrep rule: https://semgrep.dev/r?q=javascript.lang.security.audit.path-traversal.path-join-resolve-traversal
- CWE-22: https://cwe.mitre.org/data/definitions/22.html
- Node.js `path.basename`: https://nodejs.org/api/path.html#pathbasenamepath-suffix

---
_Filed by automated Semgrep triage. Source finding: Semgrep cloud, deployment `simons_personal_org`._

laboef1900:
## Summary

Closes #635.

The previous `attachmentPath(userId, pageId, filename)` helper in `attachment-handler.ts` combined `path.basename` sanitisation with a generic `// nosemgrep` line so semgrep's `path-join-resolve-traversal` would stop firing. The sanitisation was effective but implicit, and the suppression hid the fact that tainted user input still flowed into `path.join`.

This PR replaces it with **`safeAttachmentPath(pageId, filename)`** — explicit, allow-list-validated, with a containment assertion as defence-in-depth.

### What changed

- **`validatePageId`** enforces `^[A-Za-z0-9_-]+$` and rejects everything else: slashes, dots, percent-encoded payloads, NUL bytes, unicode look-alikes, empty strings. Confluence DC page IDs are numeric, so the allow-list is intentionally narrow.
- **`validateFilename`** reduces input to `path.basename`, then re-checks for empty / NUL / leading dot. `..` and `.` collapse to themselves and then fail the dotfile check.
- **`safeAttachmentPath`** runs both validators, computes the path via `path.resolve(ATTACHMENT_ROOT, pageId, filename)`, and asserts the result starts with `path.resolve(ATTACHMENT_ROOT) + path.sep` — closes the `/base` vs `/base-evil` prefix-overlap hole.
- All in-file callers were updated to use the new helper. The unused `userId` parameter is gone from the path-construction surface (public functions still accept `userId` for observability).
- The generic `// nosemgrep` on line 60 (and on line 137 inside the `readAttachment` xref fallback) are gone. The two remaining suppressions sit **inside** the validated helpers, point at the specific rule ID, and explain the invariant.

### Before / after

```diff
-function attachmentPath(_userId: string, pageId: string, filename: string): string {
-  const safe = path.basename(filename);
-  // nosemgrep
-  return path.join(attachmentDir(pageId), safe);
-}
+function safeAttachmentPath(pageId: string, filename: string): string {
+  const safeFilename = validateFilename(filename);
+  const dir = attachmentDir(pageId);
+  // nosemgrep: javascript.lang.security.audit.path-traversal.path-join-resolve-traversal.path-join-resolve-traversal -- both arguments are validated above (validatePageId, validateFilename) and containment is asserted below
+  const resolved = path.resolve(dir, safeFilename);
+  if (!resolved.startsWith(ATTACHMENTS_BASE_RESOLVED + path.sep)) {
+    throw new Error('Path traversal detected');
+  }
+  return resolved;
+}
```

### Behaviour change worth reviewing

Inputs the **old permissive sanitiser** would have coerced into a \"safe-looking\" string (e.g. `'..\\..\\etc'`, `'..%2F..%2Fetc'`, `'../../../etc/passwd'`) now **reject** up front with `Error('Invalid page ID')` rather than silently producing an underscored directory name. This is the correct hardened posture for a security boundary — every existing \"path traversal prevention\" test was rewritten to assert rejection, and 27 new direct unit tests cover dotfiles, NUL bytes, unicode look-alikes, prefix overlap, and empty input via a small `__internal` test-only export.

Callers in this codebase only pass real Confluence page IDs (numeric strings) or the controlled test fixtures (`page-1`, etc.), all of which pass the new allow-list. The new validators are exercised end-to-end by `cacheAttachment`, `readAttachment`, and friends through their full test suite.

## Verification

- `npm run lint -w backend` — clean
- `npm run typecheck -w backend` — clean
- `npx vitest run src/domains/confluence/services/attachment-handler.test.ts` — **118/118 pass** (91 existing + 27 new)
- `semgrep scan --config=p/security-audit --config=p/owasp-top-ten backend/src/domains/confluence/services/attachment-handler.ts` — **0 findings**
- `semgrep scan --config=auto …` (full rule set, includes the original `javascript.lang.security.audit.path-traversal.path-join-resolve-traversal` rule that originally fired) — **0 findings**

## Test plan

- [x] Existing 91 attachment-handler tests still green
- [x] New validator tests: `validatePageId` accepts the allow-list / rejects everything else (slashes, dots, NUL, unicode look-alike, empty, etc.)
- [x] New `validateFilename` tests: basename strip, dotfile rejection, NUL rejection, empty rejection
- [x] New `safeAttachmentPath` tests: containment, prefix-overlap guard, unicode look-alike pageId, dirname-stripping
- [x] Semgrep on the file is 0 findings under both `p/security-audit + p/owasp-top-ten` and `auto`
- [x] Lint + typecheck clean

## Notes for review

- The two remaining `nosemgrep` annotations are intentional: semgrep's dataflow tracks function-parameter taint past my regex validation (it doesn't recognise `validatePageId` as a sanitiser), so the rule fires on the `path.resolve` line itself. The suppression is narrow (specific rule ID, on the exact resolve call, with a justification linking to the validator above and the containment check below). This mirrors the pattern in #634 / [PR #648](https://github.com/Compendiq/compendiq-ce/pull/648) — single audited unsafe-looking line per security boundary.
- The `__internal` export is for test-only access to the private validators. If you'd prefer keeping the module surface absolutely private, happy to move the tests into a co-located `__tests__/internals.test.ts` and use `// @ts-expect-error` to reach in — let me know.

### DIFF
--- a/backend/src/domains/confluence/services/attachment-handler.test.ts
+++ b/backend/src/domains/confluence/services/attachment-handler.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import fs from 'fs/promises';
+import path from 'path';
 import { request } from 'undici';
 import {
   syncImageAttachments,
@@ -1267,38 +1268,33 @@ describe('attachment-handler', () => {
   });
 
   describe('path traversal prevention', () => {
-    it('allows normal pageId values', async () => {
+    it('allows normal numeric pageId values', async () => {
       const client = createMockClient();
-      // A normal pageId should work without errors
       await expect(
         cacheAttachment(client, 'user-1', '12345', '/download/img.png', 'img.png'),
       ).resolves.toBeDefined();
       expect(fs.mkdir).toHaveBeenCalled();
     });
 
-    it('blocks pageId with path traversal sequences (../)', async () => {
+    // Behaviour change vs. the previous permissive sanitiser (#635):
+    // the new helper validates pageId against /^[A-Za-z0-9_-]+$/ and throws on
+    // any character outside that allow-list. Inputs that the old code would
+    // have coerced to a "safe-looking" string now reject up front.
+
+    it('rejects pageId containing path traversal sequences', async () => {
       const client = createMockClient();
-      // After sanitization, "../../../etc" becomes "_.._.._.._etc" which is safe,
-      // but the key protection is the resolved path check
       await expect(
         cacheAttachment(client, 'user-1', '../../../etc/passwd', '/download/x.png', 'x.png'),
-      ).resolves.toBeDefined();
-      // Verify the directory created does NOT contain literal ".."
-      const mkdirCall = vi.mocked(fs.mkdir).mock.calls[0][0] as string;
-      expect(mkdirCall).not.toContain('..');
+      ).rejects.toThrow('Invalid page ID');
     });
 
-    it('sanitizes URL-encoded traversal (..%2F)', async () => {
+    it('rejects URL-encoded traversal in pageId', async () => {
       const client = createMockClient();
-      // "..%2F..%2Fetc" — dots and slashes are stripped; the %2F literal chars
-      // are harmless at the filesystem level but dots are still removed.
+      // `%` is outside the allow-list, so a payload like "..%2F..%2Fetc" is
+      // rejected before it ever reaches path.resolve.
       await expect(
         cacheAttachment(client, 'user-1', '..%2F..%2Fetc', '/download/x.png', 'x.png'),
-      ).resolves.toBeDefined();
-      const mkdirCall = vi.mocked(fs.mkdir).mock.calls[0][0] as string;
-      // Dots and slashes stripped — no traversal sequences remain
-      expect(mkdirCall).not.toContain('..');
-      expect(mkdirCall).not.toContain('/.');
+      ).rejects.toThrow('Invalid page ID');
     });
 
     it('rejects empty pageId', async () => {
@@ -1308,20 +1304,15 @@ describe('attachment-handler', () => {
       ).rejects.toThrow('Invalid page ID');
     });
 
-    it('sanitizes backslashes and dots in pageId', async () => {
+    it('rejects pageId with backslashes or dots', async () => {
       const client = createMockClient();
       await expect(
         cacheAttachment(client, 'user-1', '..\\..\\etc', '/download/x.png', 'x.png'),
-      ).resolves.toBeDefined();
-      const mkdirCall = vi.mocked(fs.mkdir).mock.calls[0][0] as string;
-      // Both dots and backslashes are replaced, no traversal possible
-      expect(mkdirCall).not.toContain('..');
-      expect(mkdirCall).not.toContain('\\');
+      ).rejects.toThrow('Invalid page ID');
     });
 
     it('rejects pageId that is only dots and slashes', async () => {
       const client = createMockClient();
-      // "../../.." becomes all underscores then trimmed — empty after trim
       await expect(
         cacheAttachment(client, 'user-1', '../../..', '/download/x.png', 'x.png'),
       ).rejects.toThrow('Invalid page ID');
@@ -1329,18 +1320,12 @@ describe('attachment-handler', () => {
 
     it('rejects pageId that is only slashes', async () => {
       const client = createMockClient();
-      // "///" becomes "___" then trimmed to empty
       await expect(
         cacheAttachment(client, 'user-1', '///', '/download/x.png', 'x.png'),
       ).rejects.toThrow('Invalid page ID');
     });
 
-    it('rejects pageId that is only dots ("...") — was masked by the old duplicate-in-char-class regex', async () => {
-      // Regression for #230: the previous regex /[/\\..]+/g had a duplicate `.`
-      // in the character class. The behaviour was identical (Zod-unrelated
-      // no-op), but leaving the bug in meant anyone editing the pattern later
-      // could misread intent. After the fix to /[/\\.]+/g, "..." still collapses
-      // to a single "_" which then trims to empty → "Invalid page ID".
+    it('rejects pageId that is only dots', async () => {
       const client = createMockClient();
       await expect(
         cacheAttachment(client, 'user-1', '...', '/download/x.png', 'x.png'),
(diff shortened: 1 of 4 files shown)


laboef1900 (commented):
## Independent review

### Verdict
Approve with nits — one behaviour-change concern worth a follow-up

### Security correctness
This is a real upgrade, not a finding suppression. The previous code relied on `path.basename` + a generic `// nosemgrep` annotation; the new helper enforces an allow-list at the boundary (`PAGE_ID_PATTERN = /^[A-Za-z0-9_-]+$/`), then asserts containment with `startsWith(ATTACHMENTS_BASE_RESOLVED + path.sep)`. The `+ path.sep` is the right fix for the classic `/base` vs `/base-evil` prefix-overlap hole, which the old code did NOT close (it used `startsWith(path.resolve(ATTACHMENTS_BASE))` without the trailing separator). That's a genuine latent bug being fixed here, not just a refactor.

The two remaining `nosemgrep` annotations are appropriately scoped (specific rule ID, narrow line, justification linking to the validator) — Semgrep's dataflow correctly can't infer regex-validation as a sanitiser, so the suppression is necessary and correct.

### Test coverage
Strong. 27 new direct unit tests via `__internal` cover the right adversarial inputs: NUL bytes, unicode look-alikes (Cyrillic "р" vs Latin "p"), URL-encoded payloads, traversal collapse via `path.basename('../..') === '..'`, dotfiles, prefix-overlap containment. The behaviour-change tests in the existing suite were updated honestly (now assert `rejects.toThrow('Invalid page ID')` instead of swallowing). One minor gap: no explicit test for the `+ path.sep` containment fix (the case where attempted output would be `/base-evil/...`). The unicode-pageId test exercises it indirectly via validator rejection, but the resolve-level containment check itself isn't directly hit because no valid input would ever cross the validator. Acceptable, but a single `it('rejects resolved path that escapes via sibling prefix', ...)` using `__internal.safeAttachmentPath` would lock in the defence-in-depth assertion.

### Concerns
- **Behaviour change for filename starting with `.`** — the PR body honestly flags the page-ID behaviour change (coerce → throw), but doesn't call out the analogous filename one. `validateFilename` now throws `Error('Invalid filename')` for any name starting with `.` (`.env`, `.htaccess`, even `.gitkeep`). The previous code would have `path.basename`-ed and proceeded; on cache miss `fs.readFile` returned `null` and the route handler at `backend/src/routes/confluence/attachments.ts:124` produced a clean 404. **Now**, because `readAttachment` doesn't catch the throw and the route handler at line 124 doesn't wrap the call, a request like `GET /api/attachments/12345/.something` will throw inside `validateFilename` and bubble up as a Fastify default 500, not a 404. Real-world impact is probably tiny (Confluence attachments rarely start with `.`), but it's a contract regression worth one of: (a) wrapping `readAttachment` to return `null` on `Invalid filename`, or (b) catching `Error` in `attachments.ts:124` and returning 404, or (c) at minimum, a route-level guard that 400s on bad filename input before calling into the handler. Option (a) is the smallest blast-radius fix.
- The `__internal` export is reasonable but exposes the API surface. It's clearly named and the only non-test importer would have to literally write `import { __internal } ...`, which would stand out in code review. Acceptable. If you want to harden further, the suggested `// @ts-expect-error` reach-in pattern would work, but I'd rather have explicit `__internal` than a comment that future-someone removes.

### Nits (won't block)
- `validateFilename` rejects empty / NUL / leading-dot but does not impose a max-length. Filesystems typically cap at 255 bytes; an attacker-controlled extremely-long filename would currently get past validation and fail at `fs.writeFile` with an opaque ENAMETOOLONG. Tightening to a max-length (e.g. 255) would convert that into a clean `Invalid filename` throw — but again, route handlers should catch.
- The two `nosemgrep` comments use the doubled rule path (`...path-join-resolve-traversal.path-join-resolve-traversal`) — that's actual semgrep rule-pack syntax, not a typo, but reads as one.
- `ATTACHMENTS_BASE_RESOLVED` is computed at module load. If `ATTACHMENTS_DIR` env var is changed at runtime (rare but possible in tests), the resolved base goes stale. Worth a comment that it's intentionally evaluated once.

laboef1900:
Addressed in f2997fb8 — wrapped the `readAttachment` call site in the GET route handler so the three validator messages (`Invalid page ID`, `Invalid filename`, `Path traversal detected`) map to a clean 404 with a warn-level log, restoring the pre-PR contract. Non-validation errors (e.g. `EACCES`) still rethrow to 500 so infrastructure issues aren't masked.

Added 6 route-level negative tests covering each path: pageId with `..`, encoded slash, dotfile filename (`.htaccess`), NUL byte (`foo%00.txt`), resolved-path escape, and a generic-error rethrow guard. All 40 tests in `attachments.test.ts` pass.

Scoped to the GET route only — the PUT route's writeAttachmentCache shares the same throw shape but the reviewer's regression concern was the read-path 404→500 change, so kept the diff focused. Happy to add the PUT-side wrapping in a follow-up if you'd prefer it bundled here.

laboef1900:
Confirmed fix landed in tree. Label updated PR:Warning → PR:Approved.

laboef1900 (commented):
## Re-review (follow-up commit f2997fb8)

### Original concern
> "a request like `GET /api/attachments/12345/.something` will throw inside `validateFilename` and bubble up as a Fastify default 500, not a 404."

### Fix verification
- **All call sites wrapped: yes.** `readAttachment` (the primary entry, previous `attachments.ts:124`) is now wrapped in a try/catch at `attachments.ts:144-167` that maps the three validator messages to a clean 404 with a `logger.warn` and rethrows everything else. The second `safeAttachmentPath`-touching path (`fetchAndCachePageImage` at line 201) is unreachable with invalid `pageId`/`filename` because the same values are validated by `readAttachment` first — if they fail, the request 404s before reaching the on-demand fetch. No additional call sites in the route surface a validator throw to an uncaught handler.
- **New test coverage: 6 negative tests** in `attachments.test.ts:325-412` cover each documented payload class — `..` in pageId, encoded slash in pageId, `.htaccess` filename, NUL-byte filename, resolved-path escape, and a non-validation rethrow guard. Each asserts `statusCode === 404` and `toMatchObject({ message: 'Attachment not found' })`; the rethrow guard asserts `500`. Good shape.
- **Error mapping correct: yes.** 404 returned with the expected body shape, warn-level log emitted with `{ userId, pageId, filename, reason }`, non-validation errors (e.g. `EACCES`) preserved as 500 so infra issues aren't masked.
- **Scope clean.** Commit touches only the route file and its tests (39 + 90 lines). No unrelated churn.

### Verdict
Approve with a new Info-level nit. The wrap is correct and the test coverage is honest.

### Follow-up nits (won't block)
- **Error-message string coupling is brittle.** `ATTACHMENT_VALIDATION_MESSAGES` matches on EXACT strings (`'Path traversal detected'`, `'Invalid page ID'`, `'Invalid filename'`). If anyone changes a throw message in `attachment-handler.ts` later, this `Set` silently misses and the case regresses to a 500 — and the existing tests wouldn't catch it because they mock `readAttachment` to reject with literal strings rather than exercising the real validators. Suggest a sentinel error class (`class AttachmentValidationError extends Error {}`) or a shared constants module, so the link is type-checked. Low priority — current code is correct as written.
- (Carrying over from the prior review, not addressed by this commit but still worth noting:) the `validateFilename` max-length omission and the doubled-rule-path `nosemgrep` comments. Both are non-blockers.

PR #675, opened 2026-05-19
fix(ai): wire Think toggle through to upstream LLM provider

laboef1900:
## Summary

The Think toggle on the AI page was wired in the UI and forwarded `thinking: true` to `/api/llm/{ask,improve,summarize}`, but the backend parsed the field and **silently dropped it** — `streamChat()` only sent `{ model, messages, stream }` upstream. Toggling Think changed the badge to purple and did nothing else.

This PR threads the flag end-to-end:

- New `thinkingExtras(model, thinking)` helper in `openai-compatible-client.ts` maps the generic flag to provider-specific extras:
  - OpenAI o-series / gpt-5 → `reasoning_effort: 'medium'`
  - Everything else (Ollama 0.6+, vLLM / SGLang serving Qwen3, DeepSeek-R1) → `think: true` + `chat_template_kwargs: { enable_thinking: true }`
- `chat()` and `streamChat()` accept an optional `opts: { thinking?: boolean }` and merge the extras into the `/v1/chat/completions` body.
- All five LLM routes (ask, improve, summarize, generate, diagram, quality) now forward `body.thinking` to `streamChat()`.
- Contract schemas: `GenerateRequestSchema`, `GenerateDiagramRequestSchema`, `AnalyzeQualityRequestSchema` gain `thinking?: boolean` (Ask/Improve/Summarize already had it).
- Frontend: `GenerateMode`, `DiagramMode`, `QualityMode` now read `thinkingMode` from `useAiContext()` and add `thinking: true` to the request body — previously the toggle was rendered globally but ignored in those three modes.

### Cache key now folds in `thinking`

`buildLlmCacheKey` and `buildRagCacheKey` (`llm-cache.ts`) didn't include `thinking`, so a warm cache from a thinking-off request would replay when the user toggled Think on — re-introducing the same no-op bug on the second turn. Both builders now accept `options.thinking` and append a `think:1` suffix when on; `false`/absent collapse to no suffix so the hash stays stable for unaffected callers. All six route call sites pass `body.thinking` into the cache key. Expect a brief cache warm-up after rollout because pre-PR cache entries hash differently.

## What changed (files)

| Area | Files |
|------|-------|
| Backend client | `openai-compatible-client.ts` (+ test) |
| Backend cache  | `llm-cache.ts` (+ test) |
| Backend routes | `llm-{ask,improve,summarize,generate,diagram,quality}.ts` |
| Contracts | `packages/contracts/src/schemas/llm.ts` |
| Frontend modes | `{Generate,Diagram,Quality}Mode.tsx` |
| E2E | `e2e/think-toggle.spec.ts` (new) |

## Why the heuristic and not "send both"

OpenAI rejects unknown JSON fields with 400, so we can't naively send `think` + `reasoning_effort` together. The model-name heuristic (`^o[1-9]` or `gpt-5*` → reasoning, else open-model) picks the right shape per backend. Caveat in the test plan below.

## Test plan

- [x] **Unit (wire format)** — 7 new assertions in `openai-compatible-client.test.ts` capture the upstream body for qwen3, deepseek-r1, o1, o3, gpt-5, and the no-thinking baseline.
- [x] **Unit (cache key)** — 3 new assertions in `llm-cache.test.ts` lock in that `thinking` produces distinct keys and that `thinking:false` equals omitted.
- [x] **Backend** — full backend route + cache suites still green: 224/224 pass.
- [x] **Frontend** — frontend AI suite still green: 172/172 pass.
- [x] **Contracts** — 66/66 pass.
- [x] **Lint + typecheck** — clean across all four workspaces.
- [x] **End-to-end via Playwright** — `e2e/think-toggle.spec.ts` registers an admin, configures a host-side mock OpenAI capture server as the chat provider, drives the AI page in a real browser, then asserts the captured upstream body. Three cases pass against a freshly-built CE Docker stack:

  ```
  ✓ toggle OFF → no thinking extras on upstream body                                  (863ms)
  ✓ toggle ON  → think + chat_template_kwargs land upstream                           (601ms)
  ✓ toggle ON after a warm thinking-off cache → upstream still receives the call      (1.1s)
  ```

  Captured upstream body for toggle ON:
  ```json
  {"model":"qwen3:8b","stream":true,"think":true,
   "chat_template_kwargs":{"enable_thinking":true}}
  ```

## Caveat — `gpt-4o` + Think on

The heuristic routes anything that isn't `o[1-9]*` / `gpt-5*` to the open-model path. A user who configures an OpenAI provider with a `gpt-4o` model **and** enables Think will get a 400 from OpenAI (unknown `think` field). This is a user error in the strict sense — `gpt-4o` is not a reasoning model — but the failure mode is a confusing 400 rather than a no-op. Two reasonable follow-ups, not in this PR:

1. Broaden the OpenAI detection to `^gpt-` and let OpenAI ignore `reasoning_effort` on non-reasoning models. (Simpler.)
2. Add a `supportsThinking` boolean to the model record and gate the Think toggle from the UI when the assigned chat model can't use it. (More correct.)

## Known UX follow-up enabled by this PR

The streaming consumer in `openai-compatible-client.ts` reads only `delta.content`. OpenAI o-series and Ollama 0.9+ stream reasoning in `delta.reasoning_content` (separate, silently dropped — fine). **Older Ollama versions** running Qwen3 / DeepSeek-R1 inline `<think>...</think>` blocks inside `delta.content`, which this PR will start surfacing in the chat UI. `quality-worker.ts:96` already strips these in the quality flow; mirroring that on the chat path (or routing reasoning to a collapsible UI block) is a follow-up.

### DIFF
--- a/backend/src/domains/llm/services/llm-cache.test.ts
+++ b/backend/src/domains/llm/services/llm-cache.test.ts
@@ -57,6 +57,12 @@ describe('buildLlmCacheKey', () => {
     const keyUndefinedProvider = buildLlmCacheKey('model', 'system', 'content', undefined);
     expect(keyNoProvider).toBe(keyUndefinedProvider);
   });
+
+  it('should differ when thinking is toggled', () => {
+    const off = buildLlmCacheKey('model', 'sys', 'user', 'p');
+    const on  = buildLlmCacheKey('model', 'sys', 'user', 'p', { thinking: true });
+    expect(off).not.toBe(on);
+  });
 });
 
 describe('buildRagCacheKey', () => {
@@ -121,6 +127,21 @@ describe('buildRagCacheKey', () => {
     const keyNoProvider = buildRagCacheKey('model', 'question', ['doc1']);
     const keyEmptyOpts = buildRagCacheKey('model', 'question', ['doc1'], {});
     expect(keyNoProvider).toBe(keyEmptyOpts);
+  });
+
+  // Thinking-on responses must live in a separate cache namespace, otherwise
+  // a prior thinking-off answer is replayed when the user toggles Think on
+  // and the upstream LLM never sees the new request.
+  it('should differ when thinking is toggled', () => {
+    const off = buildRagCacheKey('model', 'q', ['d1'], { provider: 'p' });
+    const on  = buildRagCacheKey('model', 'q', ['d1'], { provider: 'p', thinking: true });
+    expect(off).not.toBe(on);
+  });
+
+  it('should treat thinking:false as equivalent to omitted', () => {
+    const off = buildRagCacheKey('model', 'q', ['d1'], { provider: 'p' });
+    const offExplicit = buildRagCacheKey('model', 'q', ['d1'], { provider: 'p', thinking: false });
+    expect(off).toBe(offExplicit);
   });
 });
 
--- a/backend/src/domains/llm/services/llm-cache.ts
+++ b/backend/src/domains/llm/services/llm-cache.ts
@@ -40,9 +40,11 @@ export function buildLlmCacheKey(
   systemPrompt: string,
   userContent: string,
   provider?: string,
+  options?: { thinking?: boolean },
 ): string {
   const providerSuffix = provider ? `provider:${provider}` : '';
-  return KEY_PREFIX + hashLlmInputs(model, systemPrompt, userContent, providerSuffix);
+  const thinkingSuffix = options?.thinking ? 'think:1' : '';
+  return KEY_PREFIX + hashLlmInputs(model, systemPrompt, userContent, providerSuffix, thinkingSuffix);
 }
 
 /**
@@ -68,6 +70,7 @@ export function buildRagCacheKey(
     externalUrls?: string[];
     searchWeb?: boolean;
     provider?: string;
+    thinking?: boolean;
   },
 ): string {
   const sortedIds = [...docIds].sort().join(',');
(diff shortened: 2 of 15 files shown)


laboef1900:
Addressed the cache-key blocker from the review.

**Fix** (`eb05506`):
- `buildLlmCacheKey` / `buildRagCacheKey` now accept `options.thinking` and fold a `think:1` suffix into the hash when on. `false`/absent collapse to no suffix so existing cache entries keep their keys.
- All six route call sites pass `body.thinking` into the cache key.

**New regression coverage**:
- `llm-cache.test.ts`: thinking on vs off produces different keys for both builders; `thinking: false` equals omitted.
- `e2e/think-toggle.spec.ts`: new case `toggle ON after a warm thinking-off cache → upstream still receives the thinking call`. Asks the same question twice (off, then on) and asserts the mock LLM captured **two** distinct upstream calls, the second with `think: true` + `chat_template_kwargs`.

**Verification against the running stack**:

```
✓ toggle OFF → no thinking extras on upstream body                                  (863ms)
✓ toggle ON  → think + chat_template_kwargs land on upstream body                   (601ms)
✓ toggle ON after a warm thinking-off cache → upstream still receives the call      (1.1s)
3 passed
```

Captured mock-server entries for the cache-bust test (same question, two distinct upstream calls):
```
{"tail":"Question: Tell me about cache-bust-…"}
{"tail":"Question: Tell me about cache-bust-…","think":true,"ctk":{"enable_thinking":true}}
```

Backend suite: 224/224 (+18: the new cache assertions plus the existing `llm-cache.test.ts` joining the run). Lint + typecheck clean across all workspaces.

PR #676, opened 2026-05-19
fix(ai): route Think extras by provider strictness, not model name

laboef1900:
## Summary

Follow-up to #675. The previous heuristic picked the wire shape from the **model name** alone, which got it wrong: the constraint is really whether the SERVER tolerates unknown JSON fields, not how the model is named. Two failure modes followed from that:

1. **OpenAI + non-reasoning model + Think on → 400.** A user with an OpenAI provider configured for `gpt-4o` (or `gpt-4`, `gpt-3.5-turbo`, fine-tunes…) who turned Think on got a 400 from OpenAI because the model-name regex routed the request to the open-model path (`think: true` + `chat_template_kwargs`), which OpenAI rejects.
2. **Future `o10*` would have hit the same 400** because `^o[1-9]` doesn't match it — and naming conventions for self-hosted setups (renamed models, fine-tune variants like `my-team/r1-tune:v2`, `gpt-oss:20b`) routinely break model-name heuristics.

The fix branches on the **provider**, not the model:

- `isStrictOpenAiHost(baseUrl)` matches `api.openai.com` exactly by hostname (rejects substring tricks like `api.openai.com.evil.tld` and proxy URLs that contain "openai" elsewhere).
- `thinkingExtras(baseUrl, model, thinking)`:
  - **Strict (`api.openai.com`)**: emit `reasoning_effort: 'medium'` only when the model is recognized as reasoning-capable (`o1`–`o9`, `gpt-5*`). For every other model on OpenAI, emit nothing. **The toggle becomes a silent no-op rather than a 400.**
  - **Tolerant (Ollama / vLLM / SGLang / LM Studio / TGI / any custom shim)**: always emit `think: true` + `chat_template_kwargs.enable_thinking: true`. These servers ignore unknown fields when the chat template has no thinking branch, so **any user-installed model is safe**: thinking activates if the model can reason, silently no-ops if not.

## User-facing rule

> *Toggle Think; if the model can reason, you'll see it reason; if not, you get a normal answer.* **Never an error.**

This holds regardless of what model the user installs on their LLM server.

## Files changed

| File | Change |
|---|---|
| `backend/src/domains/llm/services/openai-compatible-client.ts` | New `isStrictOpenAiHost`; `thinkingExtras` now takes `baseUrl`. Both exported under `__test_only__` for direct unit testing. |
| `backend/src/domains/llm/services/openai-compatible-client.test.ts` | Replaced model-name-only assertions with a table-driven matrix: strict × reasoning (o1/o3/o4-mini/gpt-5), strict × non-reasoning (gpt-4o/gpt-4/gpt-3.5/text-embedding), tolerant × thinking template (qwen3/deepseek-r1/magistral/gpt-oss), tolerant × non-thinking template (llama3/phi-4/custom names). Plus a discrimination test for `isStrictOpenAiHost`. |

## Test plan

- [x] **Unit (matrix)** — 36/36 in `openai-compatible-client.test.ts` (was 20). Every (provider × model) row that matters in practice is one row in the table.
- [x] **Unit (cache key + everything else)** — full LLM domain + routes suite still green: 260/260 pass.
- [x] **Lint + typecheck** — clean across all workspaces.
- [x] **End-to-end via Playwright** — re-ran the existing 3-case spec against a fresh CE Docker stack with the new build. All green — the mock server isn't `api.openai.com`, so it routes through the tolerant branch and the existing assertions still hold:

  ```
  ✓ toggle OFF → no thinking extras on upstream body                                  (844ms)
  ✓ toggle ON  → think + chat_template_kwargs land on upstream body                   (625ms)
  ✓ toggle ON after a warm thinking-off cache → upstream still receives the call      (1.1s)
  3 passed
  ```

## What this PR does NOT do

- **No UI feedback when the toggle is inert.** A user who toggles Think on with a non-thinking model still sees normal output and has no way to know "thinking didn't happen". Closing that gap requires probing model capabilities from the provider (or hard-coding a model list) and greying out the toggle — left as a future follow-up if the UX gap matters.
- **Custom OpenAI proxies that mirror strict behavior** (e.g. an Azure OpenAI deployment that also rejects unknown fields) are NOT detected; they'll be treated as tolerant and may 400 on non-reasoning models. Adding an admin-side "strict provider" checkbox is the clean fix if/when this surfaces in practice.

### DIFF
--- a/backend/src/domains/llm/services/openai-compatible-client.test.ts
+++ b/backend/src/domains/llm/services/openai-compatible-client.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, beforeAll, afterAll } from 'vitest';
 import { createServer, type Server } from 'node:http';
 import type { AddressInfo } from 'node:net';
-import { checkHealth, listModels, chat, streamChat, generateEmbedding, invalidateBreaker, type ProviderConfig } from './openai-compatible-client.js';
+import { checkHealth, listModels, chat, streamChat, generateEmbedding, invalidateBreaker, __test_only__, type ProviderConfig } from './openai-compatible-client.js';
 
 let srv: Server;
 let baseUrl: string;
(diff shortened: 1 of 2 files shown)


laboef1900:
Addressed the three review findings in `490a89e`:

1. **Azure OpenAI now detected as strict.** `isStrictOpenAiCompatibleHost` (renamed for accuracy) matches `api.openai.com` exactly OR any host ending in `.openai.azure.com`. Refactored into a `STRICT_HOSTS` set + `STRICT_HOST_SUFFIXES` array so the next strict provider is a one-line constant edit. Adversarial cases (`openai.azure.com.evil.tld`) still rejected — verified with a parameterized test row.

2. **`o1` family dropped from the reasoning regex.** Tightened `^o[1-9]` → `^o[3-9]`. `o1`, `o1-mini`, `o1-preview` shipped before `reasoning_effort` existed and would 400 on it; they now fall through to the strict-non-reasoning branch and silently no-op. `o3`, `o4`, future `o5+` still emit the parameter. Extracted `isOpenAiReasoningModel(model)` so the model regex has its own focused unit table.

3. **Logging when the toggle is inert.** `logger.debug({ baseUrl, model }, 'Think requested on a strict provider but model is not reasoning-capable — emitting no extras')` — closes the "why didn't Think do anything?" support gap without affecting normal-traffic logs (debug level).

**Coverage**: 63 tests in this file (was 36), 450 across LLM domain + all routes. Lint + typecheck clean.

Deferred from review:
- Hard-coded `reasoning_effort: 'medium'` (not configurable) — kept as-is; future PR.
- `__test_only__` export gating — kept as-is; trivial vs. the testing benefit.

PR #787, opened 2026-06-11
UX review fixes: state consistency, trash repair, degraded-state UX, system-account shielding

laboef1900:
## Summary

Fixes all 20+ findings from the 2026-06-10 usability review of the running EE stack, in five themes:

### State consistency
- Pages tree, list, detail, search, and dashboard KPIs now share one visibility predicate (`core/services/page-visibility.ts`); the sidebar can no longer show another user's private page that 404s on click (#tree route), and KPI counts match what the caller can see (embedding-status scoping). All 11 legacy inline copies of the predicate migrated to the helper; the two deliberately-divergent predicates are now commented as such.
- Cross-user cache invalidation when a page's visibility changes.

### Trash (was broken end-to-end)
- Frontend called `/api/trash` + `/api/trash/:id/restore`; backend serves `/api/pages/trash` + `/api/pages/:id/restore` — the trash UI could never show anything. Paths fixed; backend now returns `deletedBy`/`autoPurgeAt`; a daily purge for standalone pages >30 days was added to the retention job (shared constant so UI copy and purge can't drift).
- Native `confirm("…cannot be undone")` (false for a soft delete) replaced by a design-system `ConfirmDialog` with honest copy, on both the primary (ArticleRightPane) and keyboard (PageViewPage) delete paths.

### Degraded-state UX (silent failures now speak)
- 403 on AI ask renders an inline error bubble naming the missing `llm:query` permission (root cause: `streamSSE` threw status-less errors — now throws `ApiError`).
- Models fetch failure shows a retry chip instead of an eternal spinner; the page summary banner says "unavailable — LLM provider offline" instead of promising a summary that can't come; the outage banner names the actual provider with a settings link instead of hardcoded "Ollama".
- License page now surfaces expired/invalid stored keys ("expired on <date>", seats/expiry grid) instead of silently showing "Community Edition — Free"; Diagnostics says "Build edition" to disambiguate.

### Security/correctness
- `__system__` sentinel hidden from admin user list and the RBAC principal picker; protected against deactivate/delete/role-change at the service layer.
- Stale CSP hash fixed (the FOUC-prevention inline script had been silently blocked in production since the rebrand) and guarded by a prebuild check; EE bundle loading is license-gated so CE deployments make zero requests for it (console errors per page load: 3 → 0 in CE, 1 benign network line in EE).

### Copy/polish sweep
article→page terminology, en-US spelling, pluralization, KPI label wrapping, hidden empty Email column, registration confirm-password with retype-clearing, settings 404 panel, disabled-Create-Page hint, feedback widget hidden on own pages (exposes `createdByUserId`), consistent selects.

Also: Confluence DC test-container runbook (`docs/confluence-test-container.md`) and the sync e2e spec aligned to the current Spaces & Sync UI (verified green against a live Confluence DC 9.2.14).

## Test plan
- Full gates green: lint, typecheck, backend 3203 / frontend 2247 / contracts+mcp-docs 71 tests
- EE merged-build suites green: CE-under-automock 3268, overlay 1137 (with the companion EE PR's exclusion fix)
- Live Playwright verification of every headline fix against rebuilt containers (incl. real expired license, real Confluence sync, trash round-trip)
- Every commit passed an individual two-stage review (spec + quality); whole-branch final review: ready to merge

Companion PR in compendiq-ee: excludes two CE community-mode tests that are structurally false inside the merged build.

### DIFF
--- a/backend/src/core/services/data-retention-service.test.ts
+++ b/backend/src/core/services/data-retention-service.test.ts
@@ -24,8 +24,22 @@ vi.mock('./audit-service.js', () => ({
   logAuditEvent: vi.fn().mockResolvedValue(undefined),
 }));
 
-import { runRetentionCleanup, startRetentionWorker, stopRetentionWorker, RETENTION_DEFAULTS } from './data-retention-service.js';
+import {
+  runRetentionCleanup,
+  startRetentionWorker,
+  stopRetentionWorker,
+  RETENTION_DEFAULTS,
+  STANDALONE_TRASH_RETENTION_DAYS,
+} from './data-retention-service.js';
 import { logAuditEvent as mockLogAuditEvent } from './audit-service.js';
+
+// pool.query call order inside runRetentionCleanup (pending_sync_versions is
+// absent: its retention-days getter is not mocked above, so that sweep skips
+// before touching the pool):
+//   [0] audit_log  [1] search_analytics  [2] error_log
+//   [3..] ADMIN_ACCESS_DENIED batches (#264)
+//   [next] standalone trash purge (UX review)
+//   [last] page_versions
 
 describe('data-retention-service', () => {
   beforeEach(() => {
@@ -48,6 +62,10 @@ describe('data-retention-service', () => {
       expect(RETENTION_DEFAULTS.error_log).toBe(30);
       expect(RETENTION_DEFAULTS.page_versions).toBe(50);
     });
+
+    it('keeps the standalone trash window at the 30 days the Trash UI promises', () => {
+      expect(STANDALONE_TRASH_RETENTION_DAYS).toBe(30);
+    });
   });
 
   describe('runRetentionCleanup', () => {
@@ -61,6 +79,8 @@ describe('data-retention-service', () => {
         .mockResolvedValueOnce({ rowCount: 3 })
         // ADMIN_ACCESS_DENIED targeted purge (#264) — short batch signals drained
         .mockResolvedValueOnce({ rowCount: 0 })
+        // standalone trash purge
+        .mockResolvedValueOnce({ rowCount: 4 })
         // page_versions
         .mockResolvedValueOnce({ rowCount: 2 });
 
@@ -71,6 +91,7 @@ describe('data-retention-service', () => {
       expect(results.error_log).toBe(3);
       expect(results.page_versions).toBe(2);
       expect(results.audit_log_admin_access_denied).toBe(0);
+      expect(results.pages_standalone_trash).toBe(4);
     });
 
     it('uses parameterized queries for time-based tables', async () => {
@@ -78,8 +99,9 @@ describe('data-retention-service', () => {
 
       await runRetentionCleanup();
 
-      // 3 umbrella time-based + 1 ADMIN_ACCESS_DENIED targeted + 1 count-based = 5
-      expect(mockPool.query).toHaveBeenCalledTimes(5);
+      // 3 umbrella time-based + 1 ADMIN_ACCESS_DENIED targeted
+      // + 1 standalone trash purge + 1 count-based = 6
+      expect(mockPool.query).toHaveBeenCalledTimes(6);
 
       // Verify audit_log uses default 365 days
       const auditCall = mockPool.query.mock.calls[0];
@@ -102,6 +124,13 @@ describe('data-retention-service', () => {
       expect(deniedCall[0]).toContain(`action = 'ADMIN_ACCESS_DENIED'`);
       expect(deniedCall[0]).toContain('LIMIT $2');
       expect(deniedCall[1]).toEqual([90, 10_000]);
+
+      // Standalone trash purge is source-scoped and parameterized on the
+      // 30-day window — never touches Confluence-synced rows.
+      const trashCall = mockPool.query.mock.calls[4];
+      expect(trashCall[0]).toContain('DELETE FROM pages');
+      expect(trashCall[0]).toContain(`source = 'standalone'`);
+      expect(trashCall[1]).toEqual([STANDALONE_TRASH_RETENTION_DAYS]);
     });
 
     it('uses ROW_NUMBER for count-based page_versions cleanup', async () => {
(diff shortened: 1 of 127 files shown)


backend/src/core/services/data-retention-service.test.ts
```diff
@@ -102,16 +124,23 @@ describe('data-retention-service', () => {
       expect(deniedCall[0]).toContain(`action = 'ADMIN_ACCESS_DENIED'`);
       expect(deniedCall[0]).toContain('LIMIT $2');
       expect(deniedCall[1]).toEqual([90, 10_000]);
+
+      // Standalone trash purge is source-scoped and parameterized on the
+      // 30-day window — never touches Confluence-synced rows.
+      const trashCall = mockPool.query.mock.calls[4];
```
laboef1900, line 130: **[Info] Index-positional assertions are brittle.** These tests assert exact `mockPool.query` call indices (`calls[4]`, `calls[5]`) and the total call count — this PR already had to renumber several when the standalone-trash sweep was inserted before `page_versions`. The next sweep added anywhere upstream will silently shift them again. Consider matching on SQL content instead, e.g. `mockPool.query.mock.calls.find(c => String(c[0]).includes('DELETE FROM pages'))`. The documenting comment block helps, but doesn't prevent the breakage.

backend/src/core/services/data-retention-service.ts
```diff
@@ -206,6 +213,72 @@ async function runAdminAccessDeniedRetention(): Promise<number> {
   return totalDeleted;
 }
 
+/**
+ * Days a soft-deleted standalone page survives in the trash before the
+ * maintenance job hard-deletes it. The Trash UI's `autoPurgeAt` field
+ * (GET /api/pages/trash) is computed from the same constant so the promise
+ * shown to the user and the actual purge can never drift apart.
+ */
+export const STANDALONE_TRASH_RETENTION_DAYS = 30;
```
laboef1900, line 222: **[Info] Hardcoded retention window.** Unlike the sibling windows (all `RETENTION_*_DAYS`, env-configurable), this one is a fixed constant. Reasonable trade-off, since the Trash UI's `autoPurgeAt` is derived from this same constant so the promise and the purge can't drift — just flagging the divergence from the configurable pattern in case ops ever wants to tune it.

backend/src/core/services/data-retention-service.ts
```diff
@@ -206,6 +213,72 @@ async function runAdminAccessDeniedRetention(): Promise<number> {
...
+ * Returns the number of pages deleted this run. Never throws (errors are
+ * logged and the promise resolves with the count so far).
+ */
+export async function purgeExpiredStandalonePages(): Promise<number> {
+  const pool = getPool();
+  let deleted = 0;
+
+  try {
+    const { rowCount } = await pool.query(
+      `DELETE FROM pages
```
laboef1900, line 246: **[Info] Unbatched DELETE vs. its batched neighbor.** `purgeExpiredStandalonePages` issues one unbounded `DELETE`, while `runAdminAccessDeniedRetention` just above deliberately batches at 10k to avoid long lock-hold times. Daily volume of >30-day standalone trash is normally tiny so this is fine today, but if deletes can be bursty, mirror the batch loop. Also worth confirming an index on `(source, deleted_at)` backs this predicate so it isn't a seq scan on a large `pages` table.

frontend/src/features/pages/PageViewPage.tsx
```diff
@@ -144,6 +156,7 @@ export function PageViewPage() {
   const [lightboxSrc, setLightboxSrc] = useState<{ alt: string; src: string } | null>(null);
   const [drawioEditingDiagram, setDrawioEditingDiagram] = useState<string | null>(null);
   const [drawioXml, setDrawioXml] = useState<string>('');
+  const [confirmTrashOpen, setConfirmTrashOpen] = useState(false);
```
laboef1900, line 159: **[Info] Move-to-trash dialog state + copy is duplicated.** `ArticleRightPane.tsx` (~line 248) carries an identical `confirmTrashOpen` state, "Move page to trash?" dialog, and handler. It works (only one opens at a time; Radix portals to `<body>`), but the copy and success/error toasts are now copy-pasted across two components — a small shared `useMoveToTrash(id)` hook would keep them in sync.

frontend/eslint.config.js
```diff
@@ -20,6 +20,16 @@ export default tseslint.config(
       'no-console': ['warn', { allow: ['warn', 'error'] }],
     },
   },
+  {
+    // Build-time Node scripts (e.g. the CSP hash prebuild guard).
+    files: ['scripts/**/*.mjs'],
```
laboef1900, line 25: **[Info] This lint override never runs.** `npm run lint` targets `src/` only, so `scripts/**/*.mjs` (the `check-csp-hash.mjs` prebuild guard) is never linted and this block is effectively dead config. Either widen the lint glob to include `scripts/`, or drop the block.

frontend/src/shared/components/badges/ServiceStatus.tsx
```diff
@@ -52,16 +54,18 @@ export function ServiceStatus() {
...
-              ? 'LLM server is unreachable'
-              : 'Ollama server is down';
+            const label = data.llmProvider
+              ? `LLM provider "${data.llmProvider}" is unreachable`
+              : 'LLM provider is unreachable';
             newAlerts.push({
+              // id stays 'ollama' for dismissal-state compatibility
               id: 'ollama',
-              service: 'ollama',
+              service: 'LLM provider',
```
laboef1900, line 63: **[Info] Stale `id: 'ollama'` outlives its rationale.** Keeping the id for dismissal-state persistence while `service` becomes `'LLM provider'` is pragmatic and documented, but the mismatch will puzzle a future reader. Longer term, a one-time migration of the persisted dismissal key would let the id match the service name.

frontend/src/features/pages/TrashPage.tsx
```diff
@@ -50,9 +50,9 @@ export function TrashPage() {
       ) : !trashData?.items.length ? (
         <div className="nm-card flex flex-col items-center justify-center py-16 text-center" data-testid="trash-empty">
           <Trash2 size={48} className="mb-4 text-muted-foreground" />
-          <p className="text-lg font-medium">No articles in trash</p>
+          <p className="text-lg font-medium">No pages in trash</p>
           <p className="text-sm text-muted-foreground">
-            Deleted standalone articles will appear here
+            Deleted local pages will appear here
```
laboef1900, line 55: **[Info] "local" is narrower than reality.** Standalone pages aren't necessarily in a *local space*, so "Deleted local pages will appear here" slightly under-describes what lands in trash. "Deleted standalone pages…" would match the `source` value the backend returns.

laboef1900 (commented):
Inline Info-level findings from `/review` (no correctness or security blockers — these are maintainability/polish notes).

PR #999, opened 2026-07-11
fix(enterprise): re-fetch license on auth change so post-mount login enables EE (#881)

laboef1900:
Fixes #881.

EnterpriseProvider fetched /admin/license once in a useEffect([]) and never re-ran after a post-mount SPA login, so an EE admin logging in from a fresh tab stayed in the community fallback all session (and logout left stale license/ui). Fix: subscribe to the auth store's accessToken via `useAuthStore((s) => s.accessToken)` and use it as the effect dependency, so the license is re-fetched on login/logout/token-refresh. License is now set unconditionally (a logged-out 401 clears it) and ui resets when the backend is not EE. Added a focused jsdom test (context.auth-relogin.test.tsx) that drives the real zustand store: one case verifies EE surfaces flip on after a post-mount login, another verifies license/ui clear on logout. Both failed on the old code and pass on the fix; all 18 enterprise tests pass; typecheck and eslint clean.

**Test:** `frontend/src/shared/enterprise/context.auth-relogin.test.tsx`

### DIFF
--- a/docs/architecture/10-flow-enterprise-license.md
+++ b/docs/architecture/10-flow-enterprise-license.md
@@ -95,11 +95,20 @@ ATM-{tier}-{seats}-{expiryYYYYMMDD}-{licenseId}.{ed25519SignatureBase64url}
 
 ```mermaid
 flowchart LR
-    boot(["App mount"]) --> fetch[["GET /api/admin/license"]]
+    boot(["App mount +<br/>every auth change<br/>(login / logout / token refresh)"]) --> fetch[["GET /api/admin/license"]]
     fetch --> decide{edition !== 'community'<br/>AND valid}
     decide -- yes --> ee["isEnterprise = true<br/>→ show OIDC tab,<br/>license form, EE features"]
     decide -- no  --> ce["isEnterprise = false<br/>→ CE UI only"]
+    fetch -- 401 / 403 --> clear["license cleared<br/>(logout / not admin)"]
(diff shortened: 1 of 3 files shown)


laboef1900:
### Code review

Found 1 issue:

1. The license-fetch trigger changed from mount-only to mount + auth change (the `useEffect` now depends on `accessToken`), but `docs/architecture/10-flow-enterprise-license.md` still shows the fetch triggered by "App mount" only, and the PR neither updates the diagram nor flags it ([CLAUDE.md](https://github.com/Compendiq/compendiq-ce/blob/2c5ffac07024610d3ccfea1e3a9dbb4d2f06b9d5/CLAUDE.md#L20-L22) says "When a code change affects system structure (compose, domains, ESLint boundaries, table/FK migrations, auth/sync/RAG/license flows, content pipeline), update the matching `docs/architecture/*.md` in the same PR. If unsure which diagram applies, flag it in the PR description.")

Change: https://github.com/Compendiq/compendiq-ce/blob/2c5ffac07024610d3ccfea1e3a9dbb4d2f06b9d5/frontend/src/shared/enterprise/context.tsx#L75-L77

Stale diagram: https://github.com/Compendiq/compendiq-ce/blob/2c5ffac07024610d3ccfea1e3a9dbb4d2f06b9d5/docs/architecture/10-flow-enterprise-license.md#L97-L99

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1006, opened 2026-07-11
fix(search): thread requested page into semantic/hybrid enhanced query (#963)

laboef1900:
Fixes #963.

Fixed dead semantic/hybrid search pagination. The enhanced query in use-search.ts called buildUrl(mode) with no page argument (so pageNum defaulted to 1 and the `page` param was never sent) and its queryKey omitted requestedPage (so changing pages neither refetched nor requested the new page), while page/totalPages were surfaced from that enhanced response — making results past page 1 unreachable in non-keyword modes. Threaded requestedPage into buildUrl(mode, requestedPage) and added requestedPage to the enhanced queryKey. Added two focused failing-first tests: one asserting the hybrid query URL includes page=2, one asserting the enhanced query refetches page 2 on rerender. All 16 tests in the file pass; no new typecheck errors on the changed file. Committed and pushed; no PR opened per instructions.

**Test:** `frontend/src/shared/hooks/use-search.test.ts`

### DIFF
--- a/frontend/src/shared/hooks/use-search.test.ts
+++ b/frontend/src/shared/hooks/use-search.test.ts
@@ -312,6 +312,54 @@ describe('useSearch', () => {
     expect(allHaveSpaceKey).toBe(true);
   });
 
+  it('sends the requested page on the enhanced (hybrid) query URL', async () => {
+    const fetchSpy = mockFetch(
+      makeSearchResponse({ mode: 'keyword', total: 25, totalPages: 3 }),
+      makeSearchResponse({ mode: 'hybrid', hasEmbeddings: true, total: 25, totalPages: 3 }),
+    );
+
+    const { result } = renderHook(
+      () => useSearch({ query: 'kubernetes', mode: 'hybrid', page: 2 }),
+      { wrapper: createWrapper() },
+    );
+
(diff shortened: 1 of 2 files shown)


laboef1900:
### Code review

Found 1 issue:

1. Two `node_modules` symlinks with absolute local paths (`/Users/simon/...`) are committed in this PR (repo root and `frontend/`). `.gitignore` only has `node_modules/` (trailing slash matches directories, not symlinks), so git tracked them. On any other clone these are broken symlinks and will break `npm install`/CI. Please `git rm node_modules frontend/node_modules` and push; consider changing the `.gitignore` pattern to `node_modules` (no trailing slash) so symlinks are ignored too.

https://github.com/Compendiq/compendiq-ce/blob/e30259df8d3cef91d981fc8121b3886bd109a8fd/node_modules

https://github.com/Compendiq/compendiq-ce/blob/e30259df8d3cef91d981fc8121b3886bd109a8fd/frontend/node_modules

The fix itself looks correct: threading `requestedPage` into `buildUrl` and the enhanced queryKey matches the root cause in #963, and both new tests fail without the change.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1018, opened 2026-07-11
fix(a11y): make sidebar & outline tree rows keyboard-operable (#880)

laboef1900:
Fixes #880

## Summary
The sidebar page tree, the local-space (drag-and-drop) tree, and the article outline all rendered each row as a clickable `<div onClick>` that carried `focus-visible:` ring classes but had no `tabIndex`, `role`, or `onKeyDown`. Such a `<div>` is not in the tab order and can never be focused, so those ring classes were dead code and Enter/Space could never activate navigation. This change turns each row into a real keyboard-operable `role="treeitem"` widget.

## Root cause
A `<div>` with no `tabIndex` is unfocusable, so keyboard-only and screen-reader users could Tab only to the occasional expand chevron (which merely toggles expansion) and then out of the tree. There was no way to focus or activate any page title, reorder a local-space page, or reach any outline heading — a WCAG 2.1.1 (Keyboard) failure on the app's primary navigation surface. The three affected surfaces:
- `SidebarTreeView.tsx` — `SidebarTreeNode` row
- `DndLocalSpaceTree.tsx` — `DndSortableTreeNode` row
- `ArticleRightPane.tsx` — `OutlineNodeItem` row

## Fix
On each of the three row `<div>`s: add `role="treeitem"`, `tabIndex={0}`, `aria-expanded` (only when the row has children), and an Enter/Space `onKeyDown` handler that reuses the existing navigate handler. `role="treeitem"` is chosen over `role="button"` because the expand/collapse chevron is a real nested `<button>`; a button role on the row would nest interactive controls. The handler guards `e.target !== e.currentTarget` so a keydown bubbling up from the focused chevron does not double-activate the row, and calls `e.preventDefault()` on Space so it navigates instead of scrolling the page. Purely additive attributes plus a keydown handler — no change to click, mouse, or scroll behavior.

## Tests
Added a keyboard-navigation suite to each of the three sibling test files (jsdom + `@testing-library/react`):
- role `treeitem` + `tabIndex` 0 on rows
- navigation on Enter
- navigation + `preventDefault` on Space
- `aria-expanded` present on expandable rows, absent on leaves
- (sidebar) a chevron keydown does not double-navigate (target guard)

All new assertions fail before the fix (no `role="treeitem"` row exists) and pass after. Full run of the three affected test files: 132 passed. `npm run typecheck` clean; ESLint on all changed files clean.

## Docs / ADR impact
None. This is a non-structural a11y fix to existing ADR-010 UI surfaces — no change to system structure, design tokens, routing, data, or backend. CLAUDE.md rule 6 does not apply.

### DIFF
--- a/frontend/src/shared/components/article/ArticleRightPane.test.tsx
+++ b/frontend/src/shared/components/article/ArticleRightPane.test.tsx
@@ -418,6 +418,39 @@ describe('ArticleRightPane', () => {
     });
   });
 
+  // #880 (code-review follow-up): outline rows carry role="treeitem" but had no
+  // role="tree" ancestor and nested-children wrappers lacked role="group", so
+  // every treeitem was orphaned (axe-critical aria-required-parent). The outline
+  // list is now a role="tree" and each expanded heading's sub-headings live in a
+  // role="group".
+  describe('outline ARIA tree semantics (#880)', () => {
+    it('exposes the outline list as a labelled ARIA tree', () => {
+      useArticleViewStore.setState({
+        headings: [{ id: 'intro', text: 'Introduction', level: 1 }],
+      });
+      render(<ArticleRightPane />, { wrapper: createWrapper() });
+      const tree = screen.getByRole('tree');
+      expect(tree).toBeInTheDocument();
+      expect(tree.getAttribute('aria-label')).toBeTruthy();
+    });
+
+    it('wraps nested sub-headings in role="group" so nested treeitems have a valid parent', () => {
+      // A level-2 heading nests under the preceding level-1 heading; outline
+      // branches are expanded by default (collapsedIds is empty).
+      useArticleViewStore.setState({
+        headings: [
+          { id: 'intro', text: 'Introduction', level: 1 },
+          { id: 'usage', text: 'Usage', level: 2 },
+        ],
+      });
+      render(<ArticleRightPane />, { wrapper: createWrapper() });
+      expect(screen.getByText('Usage')).toBeInTheDocument();
+      const group = document.querySelector('[role="group"]');
+      expect(group).not.toBeNull();
+      expect(group!.querySelector('[role="treeitem"]')).not.toBeNull();
+    });
+  });
+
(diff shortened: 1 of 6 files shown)


laboef1900:
**Known follow-up (not a blocker):** this closes the load-bearing WCAG 2.1.1 keyboard-operability failure (rows are now `role="treeitem"`, `tabIndex=0`, Enter/Space activate). Two APG-tree niceties were deferred by the plan and remain: (1) no ancestor `role="tree"` / nested `role="group"` wrappers, so SR announcement of tree position/level is incomplete (WCAG 4.1.2/1.3.1); (2) no roving-tabindex — every row is a Tab stop rather than a single stop + arrow-key nav. Both are strict improvements-pending over the pre-fix zero-operability state; suggest a tracked follow-up for full APG-tree semantics.

laboef1900:
Follow-up commit `aead06b9` addresses code-review feedback on the ARIA semantics.

The tree rows already carried `role="treeitem"`, but no ancestor had `role="tree"` and the nested-children wrappers lacked `role="group"`, so every treeitem was orphaned — an axe-critical `aria-required-parent` violation that breaks screen-reader tree semantics. This commit gives each of the three row-list containers (SidebarTreeView, DndLocalSpaceTree, ArticleRightPane outline) `role="tree"` + a descriptive `aria-label`, and each expanded node's nested-children wrapper `role="group"`, so every treeitem now has a valid required-parent context.

Clarification: the PR's original description overstated the reorder fix. This work delivers **valid ARIA tree/group semantics + Enter/Space activation** only. **Keyboard reorder** and **full roving-tabindex / arrow-key navigation** are intentionally out of scope here and remain a tracked follow-up (epic #856). The existing treeitem/tabIndex/onKeyDown logic is unchanged.

Each of the three sibling test files gained additive assertions that (a) the list container exposes `role="tree"` and (b) an expanded parent's children wrapper exposes `role="group"`. All 138 tests pass; typecheck and eslint are clean.

Issue #765, opened 2026-06-10
Bug: AI Improve drops page structure (Confluence row/column layouts + labels) — structured content lost in the Markdown round-trip


laboef1900:
## Summary

When an article is sent through **AI Improve** and the result is applied, the page's **layout structure (rows/columns)** and **labels** are lost — a multi-column Confluence layout collapses into a single column of plain content. We want AI Improve to **preserve all page styling/structure**, changing only the prose.

Root cause: AI Improve serializes the page to **Markdown** to send to the LLM and rebuilds it from Markdown on the way back. Confluence layout containers have **no Markdown representation and no preservation token**, so they are flattened and cannot be reconstructed — even though the converter already has a lossless mapping for them at the Confluence⇄HTML boundary.

## Steps to reproduce
1. Open a Confluence-synced article that uses a **multi-column layout** (Confluence "page layout" with 2+ columns) and/or page **labels**.
2. Open the AI panel → **Improve** → run any improvement (e.g. *Clarity* / *Grammar*).
3. **Accept** the diff.
4. Re-open the page (and/or check it in Confluence).

**Expected:** the row/column layout, labels, and other Confluence formatting are preserved; only the prose changes.

**Actual:** the layout collapses to a single column; labels are gone.

## Root cause — layout loss (confirmed)

AI Improve round-trips content through Markdown:

`body_html` → `htmlToMarkdown()` → LLM (Markdown) → `markdownToHtml()` → `htmlToConfluence()` → Confluence storage

The converter (`backend/src/core/services/content-converter.ts`) has a **complete, lossless layout mapping at the Confluence⇄HTML boundary**:

- `confluenceToHtml()` maps `ac:layout` / `ac:layout-section[ac:type]` / `ac:layout-cell` → `div.confluence-layout` / `div.confluence-layout-section[data-layout-type]` / `div.confluence-layout-cell` (`content-converter.ts:422-443`).
- `htmlToConfluence()` reconstructs the `ac:layout*` elements from exactly those divs (`content-converter.ts:766-784`).

But the **Markdown boundary is lossy**:

[...]

PR #1019, opened 2026-07-11
fix(editor): register Confluence macro placeholders in edit-mode schema (#857)

laboef1900:
Fixes #857

## Summary
The edit-mode TipTap editor silently dropped several Confluence macro placeholders on open, so saving a synced page permanently deleted those macros from the Confluence page. This registers the missing nodes in the editor's ProseMirror schema so every placeholder the backend emits survives a load → serialize round-trip.

## Root cause
`Editor.tsx`'s `useEditor` extension list omitted nodes for macros the backend's `confluenceToHtml` writes into `body_html`:
- `Panel`, `ConfluenceToc`, and `UnknownMacro` existed in `article-extensions.ts` and were registered in the read-only `ArticleViewer`, but **not** in the editor.
- There were **no nodes at all** for `span.confluence-jira-issue`, `div.confluence-include-macro`, `div.confluence-labels-macro`, or `span.confluence-user-mention`.

ProseMirror's DOM parser skips elements with no matching parse rule and keeps only their children, so opening a page unwrapped these wrappers to bare paragraphs/text. On save, `editor.getHTML()` omitted them and `htmlToConfluence(...)` → `client.updatePage(...)` overwrote the Confluence page without the macros — the exact #765 data-loss hazard the converter comment warns about.

## Fix (frontend only — reverse converters already exist and are correct)
- Added four pass-through nodes to `article-extensions.ts` mirroring the existing `ConfluenceStatus`/`ConfluenceChildren` pattern: `ConfluenceJiraIssue`, `ConfluenceUserMention` (inline atoms), `ConfluenceIncludeMacro`, `ConfluenceLabelsMacro` (block atoms). Each preserves its class + lowercased `data-*` params + visible label so the reverse pass matches byte-for-byte.
- Gave `UnknownMacro` a `macroName` attribute (`data-macro-name`) and `ConfluenceToc` its `data-*` param attributes so the macro name / TOC params round-trip.
- Registered `Panel`, `ConfluenceToc`, `UnknownMacro`, and the four new nodes in the edit-mode extension list (`Editor.tsx`); registered the four new nodes in `ArticleViewer.tsx` for read-mode consistency.

## Tests
New `Editor.test.tsx` case renders a realistic `body_html` fixture (panel / toc / jira / include / labels / unknown-macro / mention) and asserts `editor.getHTML()` still contains every wrapper + its params.
- **Before:** fails — output is unwrapped to `<p>hi</p><p>[Table of Contents]</p><p>[JIRA: KEY-1]</p>…`.
- **After:** passes. Full article suite green (438 tests). `typecheck` and `eslint` clean.

## Docs / ADR impact
Aligns with ADR-003. Added a round-trip note to `docs/architecture/11-content-pipeline.md` (source-of-truth for the Confluence⇄HTML macro round-trip) stating the editor's ProseMirror schema must stay in sync with the converter's placeholder set, and that new placeholders must be registered in both `Editor.tsx` and `ArticleViewer.tsx` (per CLAUDE.md rule 6). No backend/migration change.


laboef1900:
**Note on unknown-macro round-trip (resolves the "only partial" concern):**

This PR fixes the **editor/read** side — it registers `UnknownMacro` (and the other placeholder nodes) in the edit-mode ProseMirror schema so `div.confluence-macro-unknown` + `data-macro-name` survive `getHTML()` instead of being flattened to text and deleted on save.

The **backend write-back** side — `htmlToConfluence` reconstructing the original `ac:structured-macro` from `div.confluence-macro-unknown` — is handled by the sibling PR #1021… correction: **#1015 (issue #865)**, which adds exactly that reverse handler (`ac:name` from `data-macro-name`, generic `ac:parameter` restoration, and protectMedia freezing for the AI-Improve path).

Together, #857 (this PR) + #865 (#1015) give unknown macros a full, lossless Confluence round-trip. Recommend reviewing/merging them as a pair. Neither is a regression on its own: this PR already stops the flatten-to-text data loss even before #865 lands.

